PR #12, opened 2019-01-30
Resolve Python 3.5 issue

muizzk:


MuellerSeb:
The issue comes from cibuildwheel and pip.
See the following:
https://github.com/joerick/cibuildwheel/issues/122
https://github.com/pypa/pip/pull/6215

PR #17, opened 2019-04-17
Print compiler version

cotsog:


MuellerSeb:
Thank you for your support!

Issue #40, opened 2019-11-07
Dropping Python 2.7 and 3.4 support


MuellerSeb:
Since Python 2 will reach it's EOL on [**1. Jan 2020**](https://www.python.org/doc/sunset-python-2/), we will drop support for it, too.

The next minor release (v1.2) will then be **Python 3 only**.

Since TravisCI supports Windows but only for Python 3 we can then drop the use of Appveyor and stick to one CI system.

MuellerSeb:
See here: https://github.com/python3statement/python3statement.github.io/pull/272

LSchueler:
Python 3.4 has already reached its EOL on 2019-03-18.

We should drop the support of that version too.

MuellerSeb:
Now we have to fulfill our promise:
https://python3statement.org/
:blush:

MuellerSeb:
We should add this line in ``setup.py``:
```python
python_requires='>=3.5',
```
See here:
https://packaging.python.org/guides/distributing-packages-using-setuptools/#python-requires

LSchueler:
Note for self:
remove "Python 2" classifier from setup.py too whenn dropping support.

MuellerSeb:
Done with #43

PR #43, opened 2019-11-18
Only support Python >= 3.5

LSchueler:
As discussed in issue #40, I've now added the hard Python version requirements for the currently supported versions to the setup.py
Next commits, we can change the line to
`python_requires=">=3.5.*",`


LSchueler:
As the Window testing on appveyor will be replaced by TravisCI, I won't bother with the failed appveyor builds.

setup.py
```diff
@@ -40,19 +40,6 @@ def find_version(*file_paths):
...
-class _TemporaryDirectory:
-    def __enter__(self):
-        self.dir_name = tempfile.mkdtemp()
-        return self.dir_name
-
-    def __exit__(self, exc_type, exc_value, traceback):
-        shutil.rmtree(self.dir_name)
-
-
-TemporaryDirectory = getattr(
```
MuellerSeb: You have to import TemporaryDirectory from tempfile now or use it by ``tempfile.TemporaryDirectory``

LSchueler: ups, thanks

MuellerSeb:
The diff for ``.travis.yml`` makes no sense. They are identical on both branches ?!?!

PR #51, opened 2019-12-10
Update cython

LSchueler:
I made a lot of changes to the Cython implementation of the variogram estimators. The code is a lot cleaner now and the two major changes are

- Two types of estimator functions can now be used
- the code for the structured variogram estimation is now unified for the different spatial dimensions

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -180,6 +180,7 @@ CY_MODULES.append(
     Extension(
         "gstools.variogram.estimator",
         [os.path.join("gstools", "variogram", "estimator.pyx")],
+        language="c++",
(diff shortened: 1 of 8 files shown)


setup.py
```diff
@@ -180,6 +180,7 @@ class MPDistribution(Distribution):
     Extension(
         "gstools.variogram.estimator",
         [os.path.join("gstools", "variogram", "estimator.pyx")],
+        language="c++",
```
MuellerSeb: Why only the estimator.pyx? What does that do?

MuellerSeb: Ahh using libcpp. ok.

Issue #36, opened 2019-10-22
Ordinary Kriging: help please?


banesullivan:
So I'm trying to get up and running with this software, and I'm finding it rather difficult to produce a dirty ordinary kriging result. I learned geostats/kriging using SGeMS and I'm realizing much of what I learned is difficult to transfer here (so far)... I'm curious if someone could help me get some quick/dirty results with a dataset that I've used in SGeMS before. 

## The Example

I believe my Gaussian model is probably problematic

```py
import numpy as np
from gstools import Exponential, Gaussian, krige
import pyvista as pv

# Load the data
probes = pv.read("probes.vtp")
model_space = pv.read("model_space.vtr")

model = Gaussian(dim=3, var=10, len_scale=500)
cond_pos = probes.points.T # gstools convention
# conditional values - what i want to interpolate
field = probes['temperature (C)']
# Create the kriging model
krig = krige.Ordinary(model, cond_pos=cond_pos, cond_val=field)

krige_field, krige_var = krig((model_space.points.T))

model_space["krige_field"] = krige_field
model_space["krige_var"] = krige_var

model_space.contour().plot(scalars="krige_field" )
```

![download](https://user-images.githubusercontent.com/22067021/67339748-05994980-f4e9-11e9-9eaa-c8810620a668.png)

I think I am running into trouble with the `len_scale` option in the Gaussian model. GSTools enforces a limit of 1000 on that but for this dataset, I may need a higher `len_scale`?

If you all were to use this given dataset to produce an example showcasing what GSTools can do, what would that be? This is all in an effort to help me build out how PyVista might interface with GSTools

## Data

I have sparse, subsurface temperature probe data and want to krige a volume of these data.

[data.zip](https://github.com/GeoStat-Framework/GSTools/files/3758878/data.zip)

## "Truth"

I'm working towards results that look more like this (prodcued in SGeMS):

<img width="580" alt="Screen Shot 2019-10-22 at 4 29 33 PM" src="https://user-images.githubusercontent.com/22067021/67339798-22ce1800-f4e9-11e9-87c4-8841c4199ddf.png">

LSchueler:
Without having had a more in depth look at your problem yet and if the length scale bound is the problem, you can do following for example:

```
model = Gaussian(dim=3, var=10)
model.set_arg_bounds(len_scale=[0, 10000])
model.len_scale = 2000
```

LSchueler:
[...]

PR #57, opened 2020-01-11
Move examples to sphinx-gallery

banesullivan:
These changes moves all of the examples to work with [Sphinx Gallery](https://sphinx-gallery.github.io/stable/index.html) (SG). SG provides a way to show off your examples with really nice formatting and incorporating the examples as test cases to make sure the example in your docs are always in working state when a user tries them out.

Previously, the examples were in two places, in scripts under the `examples` directory and in code blocks within the documentation pages. Having the examples in two places can lead to discrepancies and maintenance issues, these changes consolidate the examples into a single place (doc and code) and provide a way to test all of the examples.

-----

I would highly recommend that you all switch from readthedocs.org to Travis and GitHub pages for building and deploying your docs. Readthedocs build environment is very limited and you will not be able to showcase any 3D examples with PyVista if you stay on readthedocs. Also, the whole purpose of this is to have your examples all in one place where they can be tested and incorporated into the documentation - Sphinx-gallery will run each example when building the docs and that is a great build status check to incorporate in with GitHub status checks (readthedocs does not support that)


banesullivan:
In it's current state, everything should build on readthedocs fine, but if you want to show add the example in #36, then you'll have to switch to using GitHub pages with a service like [`doctr`](https://github.com/drdoctr/doctr)

banesullivan:
Also, it looks like there are a ton of images in the `docs/source/pics` directory... most of these are now automatically generated by sphinx-gallery during the docs build, so they can be deleted.

banesullivan:
It appears the builds are failing because you all upload to PyPI on every build (and I don't have privileges to do that)?

banesullivan:
the base has been changed to `develop`

docs/source/conf.py
```diff
@@ -247,3 +248,47 @@ def setup(app):
...
+        "examples/00_intro/",
+        "examples/01_random_field/",
+        "examples/02_cov_model/",
+        "examples/03_variogram/",
+        "examples/04_vector_field/",
+        "examples/05_kriging/",
+        "examples/06_conditioned_fields/",
+        "examples/07_transformations/",
+    ],
+    # Patter to search for example files
```
MuellerSeb: Typo: "Pattern"

docs/source/conf.py
```diff
@@ -247,3 +248,47 @@ def setup(app):
...
+    "filename_pattern": r"\.py",
+    # Remove the "Download all examples" button from the top level gallery
+    "download_all_examples": False,
+    # Sort gallery example by file name instead of number of lines (default)
+    "within_subsection_order": FileNameSortKey,
+    # directory where function granular galleries are stored
+    "backreferences_dir": None,
+    # Modules for which function level galleries are created.  In
+    "doc_module": "gstools",
+    #"image_scrapers": ('pyvista', 'matplotlib'),
```
MuellerSeb: Why these comments?

docs/source/tutorials.rst
```diff
@@ -1,17 +1,22 @@
...
 =================
 
 In the following you will find several Tutorials on how to use GSTools to
 explore its whole beauty and power.
 
+
 .. toctree::
-   :maxdepth: 1
+   :includehidden:
+   :maxdepth: 3
```
MuellerSeb: Why ``:maxdepth: 3``? We think 1 is enough, as the chapters are very short in general.

docs/source/conf.py
banesullivan: For *when* you all start using PyVista for some 3D viz in the example gallery, you will need to add the pyvista image scraper. I figured I would leave this here for when that happens

PR #65, opened 2020-01-15
Add new data encapsulation

LSchueler:
2 new classes are added in order to separate data and methods more
cleanly. This is very much WIP, but I'd like to get a discussion
started before I put in too much work.
The `FieldData` class contains information about the mesh and it
holds a dict containing the individual field values, which are
defined on that mesh, e.g. "krige" and "cond".

I'd like to more clearly differentiate between a class instance of `Data` and the actual field values contained in that instance, i.e. the numpy array. Any ideas?
For a better discussion, I used several different ways to access both, like
- `field["srf"]`
- `field.values`
- `field.field`

Questions
--------------

- Should the class `Field` be renamed to `FieldMethods`?
- If a field is overwritten, with, say a new `mesh_type` or new `pos`, should only the field values be deleted or everything, including `mean`, ...?
- What should the `__call__` methods return?
   - the dictionary containing the field names and the `Data` instances
   - the `Data` instance of the default field
   - the values of the default field
- Should the `default_field` in the `Data` class always be called `"field"`, or should the name be more descriptive, like `"srf"` or `"krige"`?

TODO
---------

- [ ] cleanup
- [x] overload the `__setattr__` method in order to assign fields to each other
- [ ] create a better separation between the `Data` instances and the actual field values
- [ ] check for attributes being set to one of the Field children, which now belong to `Data`
- [ ] documentation
- [ ] get all the tests to run


MuellerSeb:
I love the approach of being able to access data by using the index operator.
To be backward-compatible we should provide the "field" attribute, which should point to a "default-field". The rest can change.

One issue I have with the approach is, that it is almost a re-implementation of a mesh-container like in meshio:
https://github.com/nschloe/meshio/blob/792dfdd6296a8858e365a005a87afa605834ce66/meshio/_mesh.py#L9

In pyevtk, the export routine "pointsToVTK" also creates a mesh, where the "cells" are simple vertices:
https://github.com/paulo-herrera/PyEVTK/blob/41d7e14e9b92909088b3e789632b1fd518b60d22/src/hl.py#L221

So in general, I would follow the definition of a VTK file, where we have:
- points (pos in our case),
- cells (vertices for unstructured, hypercubes for structured),
- point-data (scalar or vector [or tensor in VTK]) -> we are interested in this
- field-data (maybe here we can store sth like the default field name)
- cell-data (i dont think we need this)
- all *-data as dict{str: np.ndarray / single value}

The ``Data`` could also be dropped. Since the ``value_type`` could be determined from the data shape and the mesh_type. Also the mean should only be connected with the current ``Field`` class.
For the class-names I would propose:
- ``Data`` - drop
- ``FieldData`` -> ``Mesh`` (with point_data and field_data {"default_field": "field"})
- ``Field`` -> ``GSField`` (since it's the main actor in "GS"Tools) with:
  - "mean" as a field_data entry in the Mesh and a ``mean`` attribute pointing there
  - ``field`` attribute pointing to the point_data with the name of the default_field (from field_data)
- ``value_type`` -> function determining the value_type of a numpy array (taking mesh_type and pos.shape)
- within the current Field.mesh method, we could then also overwrite the underlying mesh with the given one.
- in every call routine, we could leave the "pos" parameter optional to use the already present underlying mesh if nothing was given
- if a new pos vector is given, the underlying point_data is erased

Then we could easy export meshes with meshio (at least in the unstructured case). Or we could totally move to pyvista and use it for IO and plotting, since there is no option to export a structured. (or we could just covert it to unstructured when exporting)

I guess @banesullivan would be happy about that, since we come closer to the VTK representation of a mesh.

These were my two cents ;-)

gstools/field/base.py
```diff
@@ -8,43 +8,236 @@
...
 from gstools.covmodel.base import CovModel
 from gstools.tools.export import to_vtk, vtk_export
 from gstools.field.tools import _get_select
 
-__all__ = ["Field"]
+__all__ = ["Field", "FieldData"]
 
 
-class Field:
+class Data:
```
MuellerSeb: This can be dropped. (See comment)

gstools/field/base.py
```diff
@@ -8,43 +8,236 @@
...
+        values: np.ndarray = None,
+        mean: float = 0.0,
+        value_type: str = "scalar",
+    ):
+        self.values = values
+        self.mean = mean
+        self.value_type = value_type
+
+
+class FieldData:
```
MuellerSeb: This could be called a ``Mesh`` class.

gstools/field/base.py
```diff
@@ -8,43 +8,236 @@
...
+        # TODO
+        if self.mesh_type == "unstructured":
+            pass
+        elif self.mesh_type == "structured":
+            pass
+        else:
+            raise ValueError("Unknown 'mesh_type': {}".format(mesh_type))
+
+
+class Field(FieldData):
```
MuellerSeb: This could be the ``GSField`` class, since it is our working horse.

LSchueler:
Thanks for your pretty thought out input!

I dig the name `Mesh`.

Are you suggesting to use nested `dict`s for storing something like the variance of the kriged field, like
```
m = Mesh()
m.add_field("krige_field", ...)
m.field_data["krige_field"]["var"] = 3.14
```
with the key of the `field_data` being the same as the name of the `point_data`? The alternative would be to keep the extra `Data` class, which we could of course rename to `FieldData`.

LSchueler:
As we don't have many variables belonging to auxiliary fields, we should simply add them to the `Mesh` class like `var_krige`.

LSchueler:
I just pushed a new suggestion for a combination of the two classes `Data` and `FieldData`, which is called `Mesh`. For better prototyping, I just put the class and a helper function into the unittest, along with some tests, to see how the class behaves.

banesullivan:
FYI: I plan to take a serious look at this before end of this week

LSchueler:
@banesullivan I'm looking forward to it. I guess it's enough if you only take a look at the file `tests/test_mesh.py`

MuellerSeb:
could you pull from develop, so we can test build and deploy with the new workflow?

LSchueler:
Yeah, good idea.

MuellerSeb:
Can you rename the branch?

LSchueler:
Oh... I just realised that type hints where introduced in Python 3.6 and Python 3.5 is still supported until 2020-09-13.
Appart from that, all the test cases are running, although I have not used any of the new possibilities yet.

MuellerSeb:
Hmm.. I think dropping support for py35 could be an option for the next release. Then we can get rid of all the py35 special treatment. (abstract base-class, ordered dicts,...)

LSchueler:
Yeah dropping support for old versions is always nice ;-)

Are you sure that changing the branch name will not confuse github and this PR?

MuellerSeb:
No, it is not possible: https://stackoverflow.com/questions/20007578/renaming-a-branch-while-on-pull-request

But maybe after being a draft, you could update the branch and make the concrete PR from another branch, maybe?

MuellerSeb:
Oooor you could create a branch called `data_encap` or something similar and then you could change the target branch of this PR ;-)

gstools/transform/field.py
```diff
@@ -335,12 +335,12 @@ def _zinnharvey(field, conn="high", mean=None, var=None):
         mean = np.mean(field)
     if var is None:
         var = np.var(field)
-    field = np.abs((field - mean) / np.sqrt(var))
```
MuellerSeb, line 338: Why reverting this? I fixed just it in https://github.com/GeoStat-Framework/GSTools/commit/0f24aad69d2602436b8dbeceb48df33517b2356d

I think the std-deviation needs to used there

gstools/transform/field.py
```diff
@@ -335,12 +335,12 @@ def _zinnharvey(field, conn="high", mean=None, var=None):
...
+    field = np.abs((field - mean) / var)
     field = 2 * erf(field / np.sqrt(2)) - 1
     field = np.sqrt(2) * erfinv(field)
     if conn == "high":
         field = -field
-    return field * np.sqrt(var) + mean
```
MuellerSeb, line 343: Why reverting this? I fixed just it in 0f24aad

I think the std-deviation needs to used there

gstools/field/base.py
```diff
@@ -26,33 +27,264 @@
...
+
+    def set_field_data(self, name, value):
+        """Add an attribute to this instance and add it the `field_data`
+
+        This helper method is used to create attributes for easy access
+        while using this class, but it also adds an entry to the dictionary
+        `field_data`, which is used for exporting the data.
+        """
+        setattr(self, name, value)
+        self.field_data[name] = value
```
banesullivan, line 116: IMO, you should avoid using `setattr` like this. In my experience, it can provide a lot of confusion for new user trying to figure out what are properties of the mesh versus data they actually used. It also can bring a lot of inconsistencies to workflows.

Also what about when a user sets a field with a space or newline character in the name? `setattr` allows those but that will not be accessible from a `.`

gstools/field/base.py
```diff
@@ -26,33 +27,264 @@
...
+        """:class:`numpy.ndarray`: The point data of the default field."""
+        return self.point_data[self.default_field]
+
+    @field.setter
+    def field(self, values):
+        self._check_point_data(values)
+        self.point_data[self.default_field] = values
+
+    @property
+    def value_type(self, field="field"):
```
banesullivan, line 179: You shouldn't have any arguments in the property accessor - its not usable by the end-user. perhaps this needs to be an standard method as right now it will always have `"field"` as the argument for `field`

gstools/field/base.py
```diff
@@ -26,33 +27,264 @@
...
+
+    @property
+    def mesh_type(self):
+        """:any:`str`: The mesh type of the fields."""
+        return self._mesh_type
+
+    @mesh_type.setter
+    def mesh_type(self, value):
+        """
+        Warning: setting a new mesh type deletes all previously stored fields.
```
banesullivan, line 195: There are a few places where you delete all previously stored fields, etc. I'd recommend making one method that will properly clear the Mesh object

gstools/field/base.py
LSchueler: The idea behind the set_field_data method is to be able to conveniently access attributes when using a Mesh class or one of its childs, but at the same time keep track of the field data properties which should be exported to VTK-files. That's what the self.field_data dictionary is for.

gstools/field/base.py
LSchueler: Good point :-D
Don't know how that happened.

gstools/field/base.py
LSchueler: What do you think, should "global" data i.e. `field_data` be deleted, if for example new `pos` are set?
I'm really unsure about that.

LSchueler:
@MuellerSeb I know this is your signature move, but what do you think of renaming `base.py` to `mesh.py`, as it could become more visible to people who actually want to use a `Mesh` and not the `SRF` for example.

MuellerSeb:
@LSchueler yeah, why not. The Field class can still be in `base.py`, right?

On the other hand, it could get a whole separate sub-module, where we could also place all IO-stuff, to get them out of the tools section. This could be future prove for incorporating more formats in the future (xarray, asc, ...)

LSchueler:
"The Field class can still be in base.py, right?" Are you kidding me?! :-D
Any suggestions for the name of the submodule? `data`? `mesh`?

LSchueler:
I just realised following:
In file `tests/test_condition.py` on lines 50 and 51 we have a problem
```python
field_1 = srf.unstructured(self.pos[0])
field_2 = srf.structured(self.pos[0])
```
When everything is based on a `Mesh`, we cannot mix up structured and unstructured stuff anymore.

MuellerSeb:
I think this should work. Let's have a chat about that.

MuellerSeb:
When a field is called with a new `pos` tuple, the mesh should be resetted.
For unstructured meshes, an `append` option could be interesting, for situations, where you want to iteratively generate new points without resetting the mesh.

LSchueler:
I've changed the SRF-class and especially the __call__ method to being more mesh-centric. In this concept, it doesn't make a whole lot of sense to give the position-tuple everytime an SRF is to be calculated. The pos-tuple belongs to the mesh.

To not break the backwards compatibility, I've included warnings.

LSchueler:
@MuellerSeb Please explain to me how `TestKrige.test_extdrift` works! You are using a structured grid, but you are generating srf's on unstructured grids. That's the only test, which does not work on this branch.

MuellerSeb:
> @MuellerSeb Please explain to me how `TestKrige.test_extdrift` works! You are using a structured grid, but you are generating srf's on unstructured grids. That's the only test, which does not work on this branch.

I'll have a look at it.

MuellerSeb:
@LSchueler : Found the problem. it is in the `Mesh._check_point_data` routine. The input pos tuple does not consist of flat arrays in this example. The Mesh class doesn't include any conversion of the pos-tuple like it was done before in the SRF class.

In the specific case, the `x` component of pos has a shape of `(51, 61)` and the field has a shape of `3111` (=51*61). in the checking routine this shape is compared to only the first entry of the shape of x (51) and there the error is thrown, that there is a shape missmatch.

I think, the pos tuple components should be flattened when the mesh is created (and converted to numpy arrays with floats).

MuellerSeb:
# Checklist:

- [ ] add `dim` input and attribute to mesh class
- [ ] pos tuple needs to be converted in `__init__` (depending on dim as an array shaped (dim, pos_count) )
- [ ] add `axis` input and attribute to mesh for sturctured meshes (pos then generated from axis on call)
- [ ] add a `convert` method to convert structured to unstructured meshes
- [ ] add a reset method to mesh, which takes the same arguments as init (init should call that)
- [ ] any `Field` calls where pos is given should reset the unerlying mesh with the reset method (depending on the mesh_type, pos should be interpreted as pos or axis like it was before) )
- [ ] `SRF` may always use the unstructured mesh internally (structured will be only used for variogram)... not sure about that one

LSchueler:
Thanks for compiling our chat from yesterday!

MuellerSeb:
Do we need another chat on that? Would be lovely to see this getting merged! :-)

LSchueler:
We have created such spaghetti code...

MuellerSeb:
> We have created such spaghetti code...

Oh no... now I feel guilty!

MuellerSeb:
Let's focus on `GSTools v2` with this one.

MuellerSeb:
Some new thoughts on this:
- actually I am quite fine with the way it is now: SRF/Krige/Field hold the information for the last generated field (pos + values)
- with the mesh class, we could provide a container living outside all these Field classes
- when evaluating on a mesh class, we can use the already present method: `Field.mesh` ([link](https://geostat-framework.readthedocs.io/projects/gstools/en/stable/field.base.html#gstools.field.base.Field.mesh)
- then we can provide simpler interfaces to meshio, ogs5py, pyvista, vtk, and so on
- code refactoring is also easier, since we don't have to care about multiple fields stored somewhere buried deep beneath `__get_items__` magic functions

LSchueler:
I'll close this PR for now, as GSTools has meanwhile gone into a different direction.

PR #70, opened 2020-02-19
Discrete transform

LSchueler:
I added a transform function to the collection, which takes values and optionally thresholds and applies these to an existing field, in order to create a field made up of a number of discrete values.


examples/07_transformations/02_discrete.py
```diff
@@ -0,0 +1,20 @@
...
+import numpy as np
+import gstools as gs
+
+# structured field with a size of 100x100 and a grid-size of 1x1
+x = y = range(100)
+model = gs.Gaussian(dim=2, var=1, len_scale=10)
+srf = gs.SRF(model, seed=20170519)
+srf.structured([x, y])
+# create 5 equidistanly spaced values
+discrete_values = np.linspace(np.min(srf.field), np.max(srf.field), 5)
```
MuellerSeb: The values are equidistantly, yes. But they don't need to be equally distributed in the end. I think, this could be misleading.

gstools/transform/field.py
```diff
@@ -67,6 +69,56 @@ def binary(fld, divide=None, upper=None, lower=None):
...
+    thresholds : :class:`numpy.ndarray`, optional
+        the thresholds, where the value classes are separated
+        Default: mean of the neighbouring values
+    """
+    if fld.field is None:
+        print("discrete: no field stored in SRF class.")
+    else:
+        if thresholds is None:
+            # just in case, sort the values
+            values = np.sort(values)
```
MuellerSeb: What if I want the values to BE unsorted? Why do the given values need to be in relation to the values of the given field?

gstools/transform/field.py
```diff
@@ -67,6 +69,56 @@ def binary(fld, divide=None, upper=None, lower=None):
...
+            thresholds = (values[1:] + values[:-1]) / 2
+        else:
+            if len(values) != len(thresholds) + 1:
+                raise ValueError(
+                    "discrete transformation: len(values) != len(thresholds) + 1"
+                )
+            values = np.array(values)
+            thresholds = np.array(thresholds)
+            for i in range(len(thresholds)):
+                if not (values[i] <= thresholds[i] < values[i + 1]):
```
MuellerSeb: Why this restriction?

gstools/transform/field.py
```diff
@@ -67,6 +69,56 @@ def binary(fld, divide=None, upper=None, lower=None):
...
+        the thresholds, where the value classes are separated
+        Default: mean of the neighbouring values
+    """
+    if fld.field is None:
+        print("discrete: no field stored in SRF class.")
+    else:
+        if thresholds is None:
+            # just in case, sort the values
+            values = np.sort(values)
+            thresholds = (values[1:] + values[:-1]) / 2
```
MuellerSeb: I think the natural way would be to us thresholds, so that the resulting ratios between the given values are even. Maybe I don't get the aim of this transformation...

examples/07_transformations/02_discrete.py
LSchueler: You mean it's misleading for this specific example? Would " create 5 eq. spaced values for this example" be better?! Or do you prefer an example with non-eq. spaced values?

gstools/transform/field.py
LSchueler: Well, at least for the way I implemented the transformation, the values have to be monotonically increasing. I'm also not sure of how to interpret non monotonically increasing values.

gstools/transform/field.py
LSchueler: How else would be define an array of thresholds, which subdivide an array of values? - It's a bit like the nodes and the edges of a graph.

MuellerSeb: The `binary` transformation does the following:
1. get a `divide` value to select "lower" and "upper" values
2. replace the lower and upper values with given values

The divide values is unrelated to the given lower and upper values, which will be set in the field. It is only related to the input-field

gstools/transform/field.py
LSchueler: Yeah, I was also not exactly sure how to handle the kind of input arguments the best.
But at least in my use case I'm just interested in the number of "value classes". The non-arithmetic mean thresholds are just a little generalisation, which was easy to implement.

If the function would just take an array of thresholds, what values would you assign to all the field values lying in between two thresholds?

LSchueler:
Thank's for the input.

I think I've addressed 1) with my other comments.

2)
At first I also wanted to simply take an integer, but by taking an array of values, it becomes so much more flexible and I simply hope, that functions like `np.linscape` are known to our users, otherwise I've given the example.

3)
But if the values are not related to the field anymore, then this should not be a field transformation, but a new kind of field generation. If you want to extend the value range of the field, it could simply be multiplied by a constant factor before transforming it.
The `binary` transformation would also only give a globally constant field, if the `divide` keyword would be larger than `max(field)` or less than `min(field)`.

LSchueler:
Wow, yesterday I had a pretty big brain fart...
I finally got your point, removed the check and added another example illustrating the new capability.

Ready for merging?

MuellerSeb:
> Wow, yesterday I had a pretty big brain fart...
> I finally got your point, removed the check and added another example illustrating the new capability.
> 
> Ready for merging?

No problem :-)

Only thing I would argue, is the standard calculation of the threshold. I would propose the following way:
```python
from scipy.special import erfinv
n = len(values)
p = np.arange(1, n) / n  # n-1 equal subdivisions of [0, 1]
# use quantile of the normal distribution to get equal portions for each value
thresholds = fld.mean + fld.model.sill * np.sqrt(2) * erfinv(2 * p - 1)
```

Then we also don't need to sort the values.

MuellerSeb:
Or we provide both ways with a switch. Then we could both be happy

LSchueler:
I get your point in wanting to relate the thresholds to a Gaussian process. But do you think there are application cases for discrete distributions where one would want that?

At least in my use case, I'd only want to have the arithmetic mean of the given values as a threshold.

Do you have an idea of how to concisely describe the difference of the two threshold calculation methods?

MuellerSeb:
> I get your point in wanting to relate the thresholds to a Gaussian process. But do you think there are application cases for discrete distributions where one would want that?

For example to get a facies distribution, where each facies has a number and they should be equally distributed and you want them to be next to each other in the order you have given them.
 
> At least in my use case, I'd only want to have the arithmetic mean of the given values as a threshold.

Then the switch could state that. like `th_mode="arithmetic"` or `th_mode="equal"`.

> Do you have an idea of how to concisely describe the difference of the two threshold calculation methods?

One connects the given values to the field and does sth. like a nearest neighbor interpolation (dividing areas by the arithmetic mean of the given values). The other chunks the given field in equal portions and equips each area with a corresponding given value.

MuellerSeb:
If we use the `thresholds` keyword, only a minimal modification is needed:
```python
if thresholds == "arithmetic":
    # just in case, sort the values
    values = np.sort(values)
    thresholds = (values[1:] + values[:-1]) / 2
elif thresholds == "equal":
    n = len(values)
    p = np.arange(1, n) / n  # n-1 equal subdivisions of [0, 1]
    # use quantile of the normal distribution to get equal portions for each value
    thresholds = fld.mean + fld.model.sill * np.sqrt(2) * erfinv(2 * p - 1)
else:
    if len(values) != len(thresholds) + 1:
        raise ValueError(
            "discrete transformation: len(values) != len(thresholds) + 1"
        )
    values = np.array(values)
    thresholds = np.array(thresholds, dtype=float)
```

LSchueler:
Okay, I've added the possibility to add more automatic threshold calculations, but your suggestion is still pretty buggy. I'll have a look at it next week.

MuellerSeb:
I just found a bug in the Zinn&Harvey transformation, where I had the same fallacy. \sigma is not the variance but the standard deviation. That's why we have to use the root.

LSchueler:
Damn, such a minor thing I added to GSTools and so much worries with it :-)
Let's finally merge this thing!

PR #75, opened 2020-03-22
Bugfix 1.2.1

MuellerSeb:
- `ModuleNotFoundError` not present in py35.
- Fixing Cressie-Bug #76 
- Adding analytical formula for integral scales of rational and stable model
- remove `prange` from `IncomprRandMeth` summators to prevent errors on Win and macOS


gstools/covmodel/models.py
```diff
@@ -325,6 +333,9 @@ def correlation(self, r):
         r = np.array(np.abs(r), dtype=np.double)
         return np.exp(-np.power(r / self.len_scale, self.alpha))
 
+    def calc_integral_scale(self):  # noqa: D102
+        return self.len_scale * sps.gamma(1.0 + 1.0 / self.alpha)
```
LSchueler, line 337: :-) That looks a lot faster

gstools/variogram/estimator.pyx
```diff
@@ -77,7 +77,7 @@ cdef inline void normalization_cressie(
         if counts[i] == 0:
             counts[i] = 1
         variogram[i] = (
-            (1./counts[i] * variogram[i])**4 /
+            0.5 * (1./counts[i] * variogram[i])**4 /
```
LSchueler, line 80: Thanks for that!

MuellerSeb:
@LSchueler should we as well change this lines to use `range`:
https://github.com/GeoStat-Framework/GSTools/blob/a01da8ecba940e67050203c00fbaad3efe6e2e88/gstools/field/summator.pyx#L213
https://github.com/GeoStat-Framework/GSTools/blob/a01da8ecba940e67050203c00fbaad3efe6e2e88/gstools/field/summator.pyx#L248
to get rid of the openmp bug on windows and macOS?

gstools/covmodel/models.py
MuellerSeb: It is!!

PR #78, opened 2020-04-14
Variogram fitting update

MuellerSeb:
This PR introduces an updated variogram fitting routine:
- fixing sill possible now
- `loss` is now selectable for smoother handling of outliers
- r2 score can now be returned to get an impression of the goodness of fitting
- weights can be passed
- instead of deselecting parameters, one can also give fix values for each parameter

Also:
- minor doc fixes
- skip travis builds for PRs from our own repo
- better setting of bounds in CovModel

### DIFF
--- a/gstools/covmodel/fit.py
+++ b/gstools/covmodel/fit.py
@@ -173,7 +173,7 @@ def fit_variogram(
         constrain_sill = True
         sill_low = model.arg_bounds["var"][0] + model.arg_bounds["nugget"][0]
         sill_up = model.arg_bounds["var"][1] + model.arg_bounds["nugget"][1]
-        if not (sill_low <= sill <= sill_up):
+        if not sill_low <= sill <= sill_up:
             raise ValueError("fit: sill out of bounds.")
         if "var" in para_select and "nugget" in para_select:
             if model.var > sill:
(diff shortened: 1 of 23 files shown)


MuellerSeb:
@LSchueler : When you are ok with that, I would merge.

gstools/covmodel/fit.py
```diff
@@ -0,0 +1,379 @@
...
+                    + "the set nugget should be less than the given sill."
+                )
+            para_select["var"] = False
+            model.var = sill - model.nugget
+        else:
+            # deselect the nugget, to recalculate it accordingly
+            # nugget = sill - var
+            para_select["nugget"] = False
+    else:
+        constrain_sill = False
```
LSchueler, line 205: Would it make sense to outsource stuff like this to helper functions for better readability?

gstools/covmodel/fit.py
```diff
@@ -0,0 +1,379 @@
...
+    r2_score : :class:`float`, optional
+        r2 score of the curve fitting results. Only if return_r2 is True.
+
+    Notes
+    -----
+    You can set the bounds for each parameter by accessing
+    :any:`CovModel.set_arg_bounds`.
+
+    The fitted parameters will be instantly set in the model.
+    """
```
LSchueler, line 148: Great doc!

gstools/covmodel/fit.py
MuellerSeb: Done. Thanks for the reminder!

gstools/covmodel/fit.py
MuellerSeb: Thanks :smile_cat:

MuellerSeb:
> The fit_variogram function is veeery long. If you promise to refactor it some time, then I'm fine with a merge right now ;-)

Refactored. You were totally right!

PR #87, opened 2020-05-02
Implemented directional variograms for `unstructured`

TobiasGlaubach:
@MuellerSeb @LSchueler I finished implementing the angle dependent variogram calculation and implemented a few test cases. Calculation is quasi pure C as soon as the calculation arrives at the calculation heavy loops. Defaults for parameters were chosen in order to preserve original behavior and the implementation should add no computational overhead if no angles are requested.

Can one of you do a code review and look at the test cases? 

Tests passed on my system without a problem. 

```
(base) C:\Users\tobia\source\repos\GSTools>python tests/test_variogram_unstructured.py TestVariogramUnstructured.test_angles
.
----------------------------------------------------------------------
Ran 1 test in 0.003s

OK
```

Test values were generated like this:

```python
import numpy as np
import gstools as gs

x = np.random.RandomState(19970221).rand(20) * 10.0 - 5
y = np.zeros_like(x)
model = gs.Exponential(dim=2, var=2, len_scale=8)
srf = gs.SRF(model, mean=0, seed=19970221)
field = srf((x, y))

bins = np.arange(10)
bin_center, gamma = gs.vario_estimate_unstructured((x, ), field, bins)

idx = np.argsort(x)

(gamma, x[idx], field[idx])
```

The implemented test-cases correspond to these "fields":
image

with 
- left being the basis for case 1
- middle being the basis for case 2 and 3
- right being the basis for case 4 and 5.

### DIFF
--- a/gstools/variogram/variogram.py
+++ b/gstools/variogram/variogram.py
@@ -84,7 +84,7 @@ def vario_estimate_unstructured(
         for 2d supply one angle which is azimuth φ (ccw from +x in xy plane)
         for 3d supply two angles which are inclination θ (cw from +z)
         and azimuth φ (ccw from +x in xy plane)
-    angles_tol : :float
+    angles_tol : class:`float`
         the tolerance around the variogram angle to count a point as being
         within this direction from another point (the angular tolerance around
         the directional vector given by angles)
--- a/gstools/variogram/variogram.py
+++ b/gstools/variogram/variogram.py
@@ -37,6 +37,8 @@ def vario_estimate_unstructured(
     pos,
     field,
     bin_edges,
+    angles=None,
+    angles_tol=0.0872665,
     sampling_size=None,
     sampling_seed=None,
     estimator="matheron",
@@ -75,6 +77,17 @@ def vario_estimate_unstructured(
         the spatially distributed data
     bin_edges : :class:`numpy.ndarray`
         the bins on which the variogram will be calculated
+    angles : :class:`numpy.ndarray`
+        the angles of the main axis to calculate the variogram for in radians
+        angle definitions from ISO standard 80000-2:2009 
+        for 1d this parameter will have no effect at all
+        for 2d supply one angle which is azimuth φ (ccw from +x in xy plane)
+        for 3d supply two angles which are inclination θ (cw from +z)
+        and azimuth φ (ccw from +x in xy plane)
+    angles_tol : :float
+        the tolerance around the variogram angle to count a point as being
+        within this direction from another point (the angular tolerance around
+        the directional vector given by angles)
     sampling_size : :class:`int` or :any:`None`, optional
         for large input data, this method can take a long
         time to compute the variogram, therefore this argument specifies
@@ -100,7 +113,18 @@ def vario_estimate_unstructured(
     field = np.array(field, ndmin=1, dtype=np.double)
     bin_edges = np.array(bin_edges, ndmin=1, dtype=np.double)
     x, y, z, dim = pos2xyz(pos, calc_dim=True, dtype=np.double)
+    
     bin_centres = (bin_edges[:-1] + bin_edges[1:]) / 2.0
+    
+    if angles is not None:
+        angles = np.array(angles, ndmin=1, dtype=np.double)
+        if angles.size == 0:
+            angles = np.append(angles, [0,0,0])
+        elif angles.size == 1:
+            angles = np.append(angles, [0,0])
+        elif angles.size == 2:
+            angles = np.append(angles, [0])
+            
 
     if sampling_size is not None and sampling_size < len(field):
         sampled_idx = np.random.RandomState(sampling_seed).choice(
@@ -118,7 +142,7 @@ def vario_estimate_unstructured(
     return (
         bin_centres,
         unstructured(
-            field, bin_edges, x, y, z, estimator_type=cython_estimator
+            field, bin_edges, x, y, z, angles, angles_tol, estimator_type=cython_estimator
         ),
     )
 
--- a/gstools/variogram/variogram.py
+++ b/gstools/variogram/variogram.py
@@ -42,6 +42,7 @@ def vario_estimate_unstructured(
     sampling_size=None,
     sampling_seed=None,
     estimator="matheron",
+    return_counts=False
 ):
     r"""
     Estimates the variogram on a unstructured grid.
@@ -104,11 +105,16 @@ def vario_estimate_unstructured(
             * "cressie": an estimator more robust to outliers
 
         Default: "matheron"
-
+    return_counts: class:`bool`, optional
+        if set to true, this function will also return the number of data 
+        points found at each lag distance as a third return value
+        Default: False
     Returns
     -------
     :class:`tuple` of :class:`numpy.ndarray`
-        the estimated variogram and the bin centers
+        1. the bin centers
+        2. the estimated variogram values at bin centers
+        (optional) 3. the number of points found at each bin center (see argument return_counts)
     """
     # TODO check_mesh
     field = np.array(field, ndmin=1, dtype=np.double)
@@ -140,12 +146,21 @@ def vario_estimate_unstructured(
 
     cython_estimator = _set_estimator(estimator)
 
-    return (
-        bin_centres,
-        unstructured(
+    estimates, counts = unstructured(
             field, bin_edges, x, y, z, angles, angles_tol, estimator_type=cython_estimator
-        ),
-    )
+        )
+
+    if return_counts:
+        return (
+            bin_centres,
+            estimates,
+            counts
+        )
+    else:
+        return (
+            bin_centres,
+            estimates
+        )
 
 
 def vario_estimate_structured(field, direction="x", estimator="matheron"):
--- a/gstools/variogram/variogram.py
+++ b/gstools/variogram/variogram.py
@@ -38,7 +38,7 @@ def vario_estimate_unstructured(
     field,
     bin_edges,
     angles=None,
-    angles_tol=0.0872665,
+    angles_tol=0.436332,
     sampling_size=None,
     sampling_seed=None,
     estimator="matheron",
@@ -87,7 +87,8 @@ def vario_estimate_unstructured(
     angles_tol : class:`float`
         the tolerance around the variogram angle to count a point as being
         within this direction from another point (the angular tolerance around
-        the directional vector given by angles)
+        the directional vector given by angles) 
+        Default: 25°≈0.436332
     sampling_size : :class:`int` or :any:`None`, optional
         for large input data, this method can take a long
         time to compute the variogram, therefore this argument specifies
--- a/gstools/variogram/variogram.py
+++ b/gstools/variogram/variogram.py
@@ -134,8 +134,9 @@ def vario_estimate_unstructured(
         angles = np.pad(
             angles, (0, dim - angles.size - 1), "constant", constant_values=0.0
         )  # fill with 0 if too less given
-        # correct intervalls for angles (only need upper hemispheres)
-        angles = angles % np.pi
+        # correct intervalls for angles
+        angles[0] = angles[0] % (2 * np.pi)
+        angles[1:] = angles[1:] % np.pi
     elif dim == 1:
         angles = None
 
--- a/gstools/variogram/variogram.py
+++ b/gstools/variogram/variogram.py
@@ -42,7 +42,7 @@ def vario_estimate_unstructured(
     sampling_size=None,
     sampling_seed=None,
     estimator="matheron",
-    return_counts=False
+    return_counts=False,
 ):
     r"""
     Estimates the variogram on a unstructured grid.
@@ -53,7 +53,8 @@ def vario_estimate_unstructured(
        \gamma(r_k) = \frac{1}{2 N(r_k)} \sum_{i=1}^{N(r_k)} (z(\mathbf x_i) -
        z(\mathbf x_i'))^2 \; ,
 
-    with :math:`r_k \leq \| \mathbf x_i - \mathbf x_i' \| < r_{k+1}` being the bins.
+    with :math:`r_k \leq \| \mathbf x_i - \mathbf x_i' \| < r_{k+1}`
+    being the bins.
 
     Or if the estimator "cressie" was chosen:
 
@@ -62,12 +63,9 @@ def vario_estimate_unstructured(
        \left|z(\mathbf x_i) - z(\mathbf x_i')\right|^{0.5}\right)^4}
        {0.457 + 0.494 / N(r_k) + 0.045 / N^2(r_k)} \; ,
 
-    with :math:`r_k \leq \| \mathbf x_i - \mathbf x_i' \| < r_{k+1}` being the bins.
+    with :math:`r_k \leq \| \mathbf x_i - \mathbf x_i' \| < r_{k+1}`
+    being the bins.
     The Cressie estimator is more robust to outliers.
-
-    Notes
-    -----
-    Internally uses double precision and also returns doubles.
 
     Parameters
     ----------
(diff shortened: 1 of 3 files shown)


MuellerSeb:
Hey @TobiasGlaubach 
thanks for your PR! We will take a look at this ASAP.

gstools/variogram/variogram.py
```diff
@@ -75,6 +77,17 @@ def vario_estimate_unstructured(
         the spatially distributed data
     bin_edges : :class:`numpy.ndarray`
         the bins on which the variogram will be calculated
+    angles : :class:`numpy.ndarray`
+        the angles of the main axis to calculate the variogram for in radians
+        angle definitions from ISO standard 80000-2:2009
```
LSchueler, line 82: Wow, I guess you're an engineer?! :-)

gstools/variogram/variogram.py
```diff
@@ -75,6 +77,17 @@ def vario_estimate_unstructured(
...
     bin_edges : :class:`numpy.ndarray`
         the bins on which the variogram will be calculated
+    angles : :class:`numpy.ndarray`
+        the angles of the main axis to calculate the variogram for in radians
+        angle definitions from ISO standard 80000-2:2009 
+        for 1d this parameter will have no effect at all
+        for 2d supply one angle which is azimuth φ (ccw from +x in xy plane)
+        for 3d supply two angles which are inclination θ (cw from +z)
+        and azimuth φ (ccw from +x in xy plane)
+    angles_tol : :float
```
LSchueler, line 87: please change to
angles_tol : :class:`float`

tests/test_variogram_unstructured.py
```diff
@@ -217,6 +217,101 @@ def test_assertions(self):
...
+        field = np.array([
+        -1.10318365, -0.53566629, -0.41789049, -1.06167529,  0.38449961,
+        -0.36550477, -0.98905552, -0.19352766,  0.16264266,  0.26920833,
+         0.05379665,  0.71275006,  0.36651935,  0.17366865,  1.20022343,
+         0.79385446,  0.69456069,  1.0733393 ,  0.71191592,  0.71969766])
+        
+        gamma_exp = np.array([
+        0.14260989, 0.18301197, 0.25855841, 0.29990083, 0.67914526,
+        0.60136535, 0.92875492, 1.46910435, 1.10165104])
+
```
LSchueler, line 235: Could you generate these values in a more "automated" way, as you described in the PR?

tests/test_variogram_unstructured.py
```diff
@@ -217,6 +217,101 @@ def test_assertions(self):
...
+        bin_centres, gamma = vario_estimate_unstructured(
+            (xr, yr),
+            field,
+            bins,
+            angles=[np.pi/4.]
+        )
+        
+        for i in range(gamma.size):
+            self.assertAlmostEqual(gamma_exp[i], gamma[i], places=3)
+
```
LSchueler, line 315: Please split these tests into different methods. In the long run, it makes debugging so much easier.

tests/test_variogram_unstructured.py
TobiasGlaubach: I am happy to generate them automatically, however I learned to implement unit test with as little integration as possible with other components, in order to test functionality isolated, and not have the test fail because the test case generation failed. 

But thats a matter of philosophy in software development I guess. I will generate the automatically as you requested.

tests/test_variogram_unstructured.py
TobiasGlaubach: Will do *thumbs up*

gstools/variogram/variogram.py
TobiasGlaubach: will do

gstools/variogram/variogram.py
TobiasGlaubach: Yep I am :-D

gstools/variogram/variogram.py
TobiasGlaubach: for my own records: DONE

tests/test_variogram_unstructured.py
TobiasGlaubach: for my own records: DONE

tests/test_variogram_unstructured.py
TobiasGlaubach: for my own records: DONE

TobiasGlaubach:
> Thank you so much for this PR! - I like the very straightforward way of enhancing the variogram estimation and the fact that the interface doesn't break backwards compatibility.
> 
> I've added some change requests into the code. Apart from those, here are a few more comments:
> 
> * How did you come up with the default value of the `angle_tol = 5°`? I imagine that's a good value for numerical experiments, but in case of actual measurements, it might be far too small. I have no clue what a good default value would be, but I agree, that we should have one :-)
> * I think some tests for the 3d case should definitely be added.
> * I'd like to see some test cases, where the field is more "spread out", but still with very few data points. Then you can test the angle tolerances better and actually exclude some, but not all points from the variogram estimation.
> * And the most important thing: Please add yourself to the `AUTHORS.md` in the main directory. Of course only if you want to.

I will look into this tomorrow or next week. Thanks for the reply.

gstools/variogram/variogram.py
MuellerSeb: Could you stick to the angles definition/explanation of the CovModel?
https://github.com/GeoStat-Framework/GSTools/blob/2c26d749ecd8554f10579bef4ac2d1f70266b5a9/gstools/covmodel/base.py#L73

TobiasGlaubach: OK, yea that would keep things consistent. Sorry, didn't saw that before.

tests/test_variogram_unstructured.py
LSchueler: Hmm, that's a good point I never thought about. Thanks for the input.

In the end, this is your work. I'll let you decide how to set up the tests.

LSchueler:
default angles:
Without having any prior experience, I have a feeling, that 25° is a sensible value.

additional testcases:
- You could generate an anisotropic field and see, if the estimated correlation lengths are in the expected range for the main directions.
- You could combine the 3 fields you have shown in the graphs in this PR and count how many field points have been taken into account when setting the angles <45° or >=45°

I think when these test cases have been added, we are good to go.

TobiasGlaubach:
Hi, 
**Update:**
I was quite busy lately so I hadn't had time to work on this much. However I found some time to start working on this again since yesterday. Currently I am implementing test cases and I am also using the prototype for a real problem of variogram fitting (as a more realistic testing scenario). Also I found a paper (https://doi.org/10.1016/j.cageo.2018.07.011), where directional variograms are calculated from open source available data using a different tool ([SGeMs](http://sgems.sourceforge.net/)), which can give us some valid data for cross validation.

@MuellerSeb  @LSchueler Most tools give a bin tolerance for overlapping lags/bins for estimating unstructured variograms. GSTools does not seem to have this, what do you think? Should we implement this?

TobiasGlaubach:
@MuellerSeb : I checked the covariance model angles and if I checked correctly found them to be deviating from the "normal" right hand side rule (standard rotation matrices) which also align spherical coordinate definitions and standard intrinsic Tait-Brian-Angles. 

Usually, if the Z+ axis points upwards, and a 'standard' rotation is used, the rotation is counter clockwise. This leads to points on the XY plane rotating from X axis towards Y axis. Same goes for the Tait-Brian-Angles. See here:

image [Taitbrianangles.svg from wikipedia](https://en.wikipedia.org/wiki/Euler_angles#/media/File:Taitbrianangles.svg)

However the definition of the Covariance model rotates the Variogram in CCW and the major axis aligns with Y. I generated a model like this:

```python
model = gs.Gaussian(dim=3, var=2.0, len_scale=[10, 2], angles=[np.deg2rad(azim), np.deg2rad(0)])
```

And plotted is in 3D as point cloud for azim in [0,10,50,80] and found that:

1. rotation is CCW
2. Major axis aligns with Y

image

Is that intended? Or is this a bug?

MuellerSeb:
@TobiasGlaubach :
These angles always give me a headache!
I just checked it with (`np.pi/4` at different positions in `angles`):
```python
import numpy as np
import gstools as gs
model = gs.Gaussian(dim=3, var=2.0, len_scale=[8, 4, 2], angles=[0, np.pi/4, 0])
model.plot("vario_spatial")
```
I found, that the rotation in the `xz` plane (around `y`-axis) is clockwise (both others are counterclock wise, as far as I can tell)... There is a bug here:
https://github.com/GeoStat-Framework/GSTools/blob/2c26d749ecd8554f10579bef4ac2d1f70266b5a9/gstools/tools/geometric.py#L59

This line should read:
```python
return np.array(((cos, 0.0, -sin), (+0.0, 1.0, +0.0), (sin, 0.0, cos)))
```

Could you add this fix? Thanks for the sharp eye!

MuellerSeb:
But following this:
https://en.wikipedia.org/wiki/Rotation_matrix#General_rotations

it should be correct this way!

MuellerSeb:
This example also shows, that the rotation is counter-clockwise:
https://geostat-framework.readthedocs.io/projects/gstools/en/stable/examples/01_random_field/02_fancier.html#sphx-glr-examples-01-random-field-02-fancier-py

TobiasGlaubach:
> This example also shows, that the rotation is counter-clockwise:
> https://geostat-framework.readthedocs.io/projects/gstools/en/stable/examples/01_random_field/02_fancier.html#sphx-glr-examples-01-random-field-02-fancier-py

This example is 2D not 3D. Can there be a difference?

TobiasGlaubach:
> But following this:
> https://en.wikipedia.org/wiki/Rotation_matrix#General_rotations
> 
> it should be correct the way it is!

Let me check the rotations from Wikipedia and the implementation in your code. I will get back to you in a hour or so.

MuellerSeb:
> This example is 2D not 3D. Can there be a difference?

Then have a look at this:
https://geostat-framework.readthedocs.io/projects/gstools/en/stable/examples/02_cov_model/02_aniso_rotation.html#rotation-angles

There you can check, that rotation in the xy and yz plane is counter-clockwise and in the xz plane it is clockwise.

TobiasGlaubach:
Hmm... it looks like you're not the only one having problems with the angles :-/

Sorry to be such a pain, but I get different results compared to the `model.plot()` function when I plot the variogram data as scatter plot in 3D. 

e.G: 
```python
import gstools as gs
elev = 10
model = gs.Gaussian(dim=3, var=2.0, len_scale=[10, 2], angles=[np.deg2rad(0), np.deg2rad(elev)])
```
results in the following plot (after changing to y slice). 

image

When I plot the same data in 3d using scatter plots I get this plot:
image


For me the 'major' axis is y aligned and the rotation in elevation (major-Z plane after first rotation = x') is CCW.

I generated the scatter plot with this script:
https://colab.research.google.com/drive/1SYXxmYzaGAK_j9raBStBG_DibnKXy78V?usp=sharing

@MuellerSeb : can you check if I am doing something wrong when plotting before we proceed any further?

MuellerSeb:
Hey there,
I guess the problem in your code is in the line:
```python
X,Y,Z = np.meshgrid(xgrid,ygrid,zgrid)
```
That should be:
```python
X,Y,Z = np.meshgrid(xgrid,ygrid,zgrid, indexing="ij")
```
Like done here:
https://github.com/GeoStat-Framework/GSTools/blob/a01da8ecba940e67050203c00fbaad3efe6e2e88/gstools/field/tools.py#L211

This indexing is another source of headache!

TobiasGlaubach:
OK, thanks a lot and sorry for the inconvenience! I will check my angles vs the covariance model angles then. I am fairly certain, the implementations are equal.

Another question: Is GSTools for release on PyPy compiled with OpenMP support? If not, I have a implementation of estimate.pyx which implements caching for distance and angle data. In case of single thread, this can significantly speed up the calculation (especially for large bin vectors). 

I could pack it into this PR as well.

MuellerSeb:
OpenMP support can be enabled by the user:
https://geostat-framework.readthedocs.io/projects/gstools/en/stable/#pip

Most gstools routines in cython use `prange` which is using openmp.

On Conda, openmp is enabled by default. On pip we didn't want to force the user to provide openmp. What do you mean exactly by "release on pypy"?

TobiasGlaubach:
Sorry I ment PyPi --> `pip install gstools`

My version enables caching with an additional parameter otherwise it just uses the previous implementation with prange. Would you want that feature as well or keep it as simple as possible for better maintainability?

MuellerSeb:
> My version enables caching with an additional parameter otherwise it just uses the previous implementation with prange. Would you want that feature as well or keep it as simple as possible for better maintainability?

I would need to have a look at it, to decide. But it sounds interesting! ;-)
Only condition is: it shoud also work without `openmp` and the user shouldn't need to care about parallelization.

TobiasGlaubach:
I had to make a new branch and commit it first. 

Here is the full branch: 
https://github.com/TobiasGlaubach/GSTools/tree/dev_caching

Here is the caching implementation:
https://github.com/TobiasGlaubach/GSTools/blob/289f62fa64e70b8900fc202650ebc1ffe8764c7a/gstools/variogram/estimator.pyx#L243-L270

And here is an example script I wrote for all new functionalities:
https://github.com/TobiasGlaubach/GSTools/blob/dev_caching/examples/03_variogram/02_variogram_estimation_options.py

Let me know, what you think. I will implement the 3D test cases for the original PR today or tomorrow.

TobiasGlaubach:
I implemented some basic 3d testcases, as well as a fix for the estimator function. Waiting for review now.

gstools/variogram/variogram.py
LSchueler: Please add that these arguments are optional and that the angles can also be float

MuellerSeb:
Is there still some progress here? :-)

TobiasGlaubach:
Hi,
A few weeks ago I did some testing with 3d cases. But it stagnant due to reliable expected results to compare against. Since then I hadn't had much time. From next week on I will have a bit more time and hope to get this ready.

Also if anyone has some reliable test cases for 3d data, It would be greatly appreciated.

LSchueler:
Hi @TobiasGlaubach,
do you think you will have some time to finish things up on this PR? - I think your almost there ;-)
Otherwise, we could think if either @MuellerSeb or I could pull your contributions together and merge it.

TobiasGlaubach:
Hi @LSchueler Thanks for checking in on this. I am lacking some verified test cases for 3D data, but I am fairly certain the algorithms are right and the angles correspond to the angle definitions generally used in GSTools. whats your opinion on this? Shall we add a warning somewhere?

MuellerSeb:
@TobiasGlaubach why not just generate a field with gstools, so we know the properties of the given field?

MuellerSeb:
@TobiasGlaubach going through your code, I noticed two problems:

1. ATM your code only take a pair of points into account, when the first aims to the second with correct angles (example 2D):
    ```python
    cdef double phi = atan2(dy,dx)
    ```
    But we would also need the other way around (that is skipped in the variogram loop):
    ```python
    cdef double phi = atan2(-dy,-dx)
    ```
    Same for 3D with `-dx`, `-dy` and `-dz`.

2. In 3D i wonder about your formula for theta
    ```python
    cdef double theta = atan2(dz,sqrt(dx + dy))
    ```
    Shouldn't that be (acos present in libc.math):
    ```python
    cdef double theta = acos(dz / sqrt(dx ** 2 + dy ** 2 + dz ** 2))
    ```
    as described here: https://en.wikipedia.org/wiki/Spherical_coordinate_system#Cartesian_coordinates
    (squares missing and atan2 doesn't map correctly into [0, pi) for elevation)

I can take over if you want. I just realized, we don't have to care about trait-bryan angles, since we only check the variogram in one given direction. Retrieving trait brian angles from sample data should be another step.

MuellerSeb:
I did some updates:
- better pre-processing for given angles: all in the interval [0, pi) except azimuth in [0,2pi) 
- check both directions for point-pairs in empirical variogram (only one combination will occur)
- use great-circle distance in 3D to check the given angle-tolerance (solves your problem at the poles)

I got some concerns/ideas:
- It would be nice, if one also provide the direction as a vector (then angle tolerance comparison would also be easier)
- it would be nice to also provide multiple directions to check (for example in 2D checking 4 directions [N-E-S-W])
- in order to be ready for nD, vectors would be better to use as direction (rather then n-D spherical coordinates like now)
- this would allow for example to check main axes from Tait-Bryan angles

But I am happy with this first implementation. Later on, we can implement a switch to use either angles for direction or vectors.

MuellerSeb:
I'll merge this into a new branch here to take over.
I got some further ideas regarding bandwith and vectors for direction.
Will keep your angles formulation.

Thanks for everything @TobiasGlaubach !

TobiasGlaubach:
@MuellerSeb Thanks for taking over, I was on vacation for a few weeks and atm I am quite buisy with other projects at the moment. Glad to have been able to contribute to your project :-) 

keep up the good work!

PR #95, opened 2020-06-03
Krige unification

### LINKED ISSUE #79: Proposal: unify kriging routines
Since all kriging routines could be reformulated in order to use the covariance matrix instead of the variogram matrix, we could unify the kriging base class to serve as a swiss knife for kriging.

All other kriging routines are then just a shortcut with a limited input parameter list.

### LINKED ISSUE #89: Issues with Kriging values
I am performing an ordinary kriging and the results I get after performing the kriging are totally wrong. For simplicity, my grid points are also my data points. The working code is:

```
"Import libraries and modules"
import numpy as np
import gstools as gs
import tkinter as tk
from tkinter import filedialog
import pandas as pd
from scipy.spatial.distance import pdist, squareform

"Import data"
root = tk.Tk()
root.withdraw()
filepath = filedialog.askopenfilenames(parent=root, title='Choose a file')
data = pd.DataFrame()
for file in filepath:
    df = pd.read_excel(file)
    data = data.append(df)
data = data.iloc[:, :3]
data = data[~np.isnan(data).any(axis=1)]
data = np.array(data, dtype=np.float)

"Calculate semivariance"
sill = np.var(data[:, 2])
pwdist = squareform(pdist(data[:, :2]))
max_lag = np.int(np.ceil(np.max(pwdist)))
npoint = np.int(np.floor(max_lag/40)) 
firstlag = max_lag / 100
lags = np.linspace(firstlag, max_lag, npoint)
bin_center, gamma = gs.vario_estimate_unstructured((data[:,0], data[:,1]), data[:,2], lags)

"Fit theoretical variogram to data"
fit_model = gs.Stable(dim=2)
fit_model.set_arg_bounds(var=[0, sill])
fit_model.set_arg_bounds(len_scale=[0, 1000000])
fit_model.fit_variogram(bin_center, gamma, nugget=False)
ax = fit_model.plot(x_max=max_lag)
ax.plot(bin_center, gamma)
print(fit_model)

"Perform kriging"
OK = gs.krige.Ordinary(fit_model, [data[:, 0], data[:, 1]], data[:, 2])
OK.unstructured([data[:, 0], data[:, 1]])
ax = OK.plot()
ax.set_aspect("equal")
```
The semivariogram looks good (as attached below). 
GSTool_Stable Variogram

The value of my kriging should range between 0 and 400 as `data[:, 2]` ranges between this value. However, i am getting values between -3000 and 8000. 
GSTool_Unstructured_OK

 Find  attached the data to reproduce the problem

[CNDA-ESP_SIMRAD_MDS_DGPS_2000_2001.xlsx](https://github.com/GeoStat-Framework/GSTools/files/4574807/CNDA-ESP_SIMRAD_MDS_DGPS_2000_2001.xlsx)

MuellerSeb:
Closes #79 
Fixes #89 

Here comes the great unification of our `krige` submodule:

- [x] Swiss Army Knife for kriging: The `Krige` class now provides everything in one place
- [x] "Kriging the mean" is now possible with the switch `only_mean` in the call routine
- [x] `Simple`/`Ordinary`/`Universal`/`ExtDrift`/`Detrended` are only shortcuts to `Krige` with limited input parameter list
- [x] We now use the `covariance` function to build up the kriging matrix
- [x] An `unbiased` switch was added to enable simple kriging (where the unbiased condition is not given)
- [x] An `exact` switch was added to allow smother results, if a `nugget` is present in the model
- [x] An `cond_err` parameter was added, where measurement error variances can be given for each conditional point
- [x] pseudo-inverse matrix is now used to solve the kriging system (can be disabled by the new switch `pseudo_inv`), this is equal to solving the system with least-squares and prevents numerical errors
- [ ] New examples added
- [ ] New tests added

### DIFF
--- a/gstools/krige/base.py
+++ b/gstools/krige/base.py
@@ -13,9 +13,14 @@ The following classes are provided
 import collections
 import numpy as np
 
-# from scipy.linalg import inv
 from scipy.spatial.distance import cdist
-from gstools.field.tools import reshape_field_from_unstruct_to_struct
+import scipy.linalg as spl
+from gstools.field.tools import (
+    reshape_field_from_unstruct_to_struct,
+    make_anisotropic,
+    rotate_mesh,
+)
+from gstools.tools.geometric import pos2xyz, xyz2pos
 from gstools.field.base import Field
 from gstools.krige.krigesum import krigesum
 from gstools.krige.tools import (
@@ -30,7 +35,7 @@ __all__ = ["Krige"]
 
 class Krige(Field):
     """
-    A base class for kriging.
+    A Swiss Army knife for kriging.
 
     Parameters
     ----------
@@ -40,10 +45,6 @@ class Krige(Field):
         tuple, containing the given condition positions (x, [y, z])
     cond_val : :class:`numpy.ndarray`
         the values of the conditions
-    mean : :class:`float`, optional
-        mean value of the kriging field
-    ext_drift : :class:`numpy.ndarray` or :any:`None`, optional
-        the external drift values at the given cond. positions (only for EDK)
     drift_functions : :class:`list` of :any:`callable`, :class:`str` or :class:`int`
         Either a list of callable functions, an integer representing
         the polynomial order of the drift or one of the following strings:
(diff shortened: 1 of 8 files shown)


gstools/krige/base.py
```diff
@@ -122,24 +165,35 @@ def __call__(
...
+                )
+                c_slice = slice(*chunk_slice)
+                # get RHS of the kriging system
+                k_vec = self._get_krige_vecs(
+                    (x, y, z), chunk_slice, ext_drift, only_mean
+                )
+                # generate the raw kriging field and error variance
+                field[c_slice], krige_var[c_slice] = krigesum(
+                    self._krige_mat, k_vec, self._krige_cond
+                )
```
LSchueler, line 196: This loop could be refactored into its own private method for readability.

MuellerSeb: I think, I'll leave it this way, because a separate routine would have a really long call signature at this point. I can live with the 10 extra lines ;-)

MuellerSeb:
The latest commits (after 6565d55) are not listed here... I'll close this PR and open another one to see if this helps.

PR #97, opened 2020-07-15
Krige unification

### LINKED ISSUE #79: Proposal: unify kriging routines

### LINKED ISSUE #89: Issues with Kriging values

MuellerSeb:
Closes #79 
Fixes #89 

Here comes the great unification of our `krige` submodule:

- [x] Swiss Army Knife for kriging: The `Krige` class now provides everything in one place
- [x] "Kriging the mean" is now possible with the switch `only_mean` in the call routine
- [x] `Simple`/`Ordinary`/`Universal`/`ExtDrift`/`Detrended` are only shortcuts to `Krige` with limited input parameter list
- [x] We now use the `covariance` function to build up the kriging matrix
- [x] An `unbiased` switch was added to enable simple kriging (where the unbiased condition is not given)
- [x] An `exact` switch was added to allow smother results, if a `nugget` is present in the model
- [x] An `cond_err` parameter was added, where measurement error variances can be given for each conditional point
- [x] pseudo-inverse matrix is now used to solve the kriging system (can be disabled by the new switch `pseudo_inv`), this is equal to solving the system with least-squares and prevents numerical errors
- [x] New examples added
- [x] New tests added

PS: I needed to create a new PR, since there were some commits in the branch, that weren't recognized by the last PR.

### DIFF
--- a/examples/05_kriging/09_pseudo_inverse.py
+++ b/examples/05_kriging/09_pseudo_inverse.py
@@ -1,0 +1,38 @@
+r"""
+Redundant data and pseudo-inverse
+---------------------------------
+
+It can happen, that the kriging system gets numerically unstable.
+One reason could be, that the input data contains redundant conditioning points
+that hold different values.
+
+To smoothly deal with such situations, you can switch on using the pseuodo
+inverse for the kriging matrix.
+
+This will result in the average value for the redundant data.
+
+Example
+^^^^^^^
+
+In the following we have two different values at the same location.
+The resulting kriging field will hold the average at this point.
+"""
+import numpy as np
+from gstools import Gaussian, krige
+
+# condtions
+cond_pos = [0.3, 1.9, 1.1, 3.3, 1.1]
+cond_val = [0.47, 0.56, 0.74, 1.47, 1.14]
+# resulting grid
+gridx = np.linspace(0.0, 8.0, 81)
+# spatial random field class
+model = Gaussian(dim=1, var=0.5, len_scale=1)
+
+###############################################################################
+krig = krige.Ordinary(model, cond_pos=cond_pos, cond_val=cond_val)
+krig(gridx)
+
+###############################################################################
+ax = krig.plot()
+ax.scatter(cond_pos, cond_val, color="k", zorder=10, label="Conditions")
+ax.legend()
--- a/examples/05_kriging/09_pseudo_inverse.py
+++ b/examples/05_kriging/09_pseudo_inverse.py
@@ -6,8 +6,8 @@ It can happen, that the kriging system gets numerically unstable.
 One reason could be, that the input data contains redundant conditioning points
 that hold different values.
 
-To smoothly deal with such situations, you can switch on using the pseuodo
-inverse for the kriging matrix.
+To smoothly deal with such situations, you can use the pseudo
+inverse for the kriging matrix, which is enabled by default.
 
 This will result in the average value for the redundant data.
 
--- a/gstools/krige/base.py
+++ b/gstools/krige/base.py
@@ -87,13 +87,15 @@ class Krige(Field):
         kriging matrix. If `True`, this leads to more numerical stability
         and redundant points are averaged. But it can take more time.
         Default: True
-    pseudo_inv_type : :class:`int`, optional
+    pseudo_inv_type : :class:`int` or :any:`callable`, optional
         Here you can select the algorithm to compute the pseudo-inverse matrix:
 
             * `1`: use `pinv` from `scipy` which uses `lstsq`
             * `2`: use `pinv2` from `scipy` which uses `SVD`
             * `3`: use `pinvh` from `scipy` which uses eigen-values
 
+        If you want to use another routine to invert the kriging matrix,
+        you can pass a callable which takes a matrix and returns the inverse.
         Default: `1`
     """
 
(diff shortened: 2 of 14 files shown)


MuellerSeb:
I still need to update the call signatures of the different kriging methods to include the `pseudo_inv_type`.

MuellerSeb:
> I still need to update the call signatures of the different kriging methods to include the `pseudo_inv_type`.

Done. @LSchueler you can start your review ;-)

examples/05_kriging/09_pseudo_inverse.py
```diff
@@ -0,0 +1,38 @@
+r"""
+Redundant data and pseudo-inverse
+---------------------------------
+
+It can happen, that the kriging system gets numerically unstable.
+One reason could be, that the input data contains redundant conditioning points
+that hold different values.
+
+To smoothly deal with such situations, you can switch on using the pseuodo
```
LSchueler, line 9: Typo... "pseuodo"

gstools/krige/base.py
```diff
@@ -28,9 +33,12 @@
 __all__ = ["Krige"]
 
 
+P_INV = {1: spl.pinv, 2: spl.pinv2, 3: spl.pinvh}
```
LSchueler, line 36: A minor suggestion:

Why not let the methods take a function, instead of keys to this look-up-table, with a recommendation in the docstrings to use one of the following scipy-functions?
This would make things much more flexible, in case someone has a much better algorithm for computing the pseudo inverse. This will probably never happen, but at least things are not hidden behind magic numbers like 1, 2, 3.

MuellerSeb: Why not both? I wanted to bring it in line with PyKrige (https://github.com/GeoStat-Framework/PyKrige/pull/151), where I need to pass this to a cython function and I don't want to pass a python function there. From my feeling, I like switches ;-)
But for UX reasons it would be nice to be able to have the ability to pass functions.

LSchueler: "where I need to pass this to a cython function" - woah, don't go down that road!! I've been there before

https://stackoverflow.com/questions/17807359/cython-how-to-assign-a-c-function-to-a-python-cython-variable

Wow, that was actually my first question on SO.

But what do you mean with "Why not both?" - I think some of our function interfaces are already pretty... bloated, with arguments excluding each other and so on. - If you want to stay in line with PyKrige, than I'd suggest to simply stick to the lookup table. Maybe with a hint in the docstrings how this can be overwritten/ extended.

MuellerSeb: With both I mean, that it can take a number or a callable. So you have a set of default methods and the possibility to just pass a callable right there:
```python
pseudo_inv_type=your_fancy_func
```
or
```python
pseudo_inv_type=1
```

LSchueler: Okay, you like many options in the function arguments... then let's finally merge!

Issue #54, opened 2020-01-04
Lat-Lon support (geographical coordinates)


MuellerSeb:
To really earn the name "Geo"Stat-tools, it would be nice to support lat-lon coordinates in the kriging and field-generation routines.

The RandMeth routines are not able to work with latlon coordinates since it's an approximation of the fourier-transformation in euclidean space which differs from the FT on the manifold of a sphere.

A trick to prevent this problem could be to use 3D fields and evaluate the field on a sphere. For rather small correlation lengths this shouldn't be a problem but with increasing length-scales the error between the spatial distance and the great-circle distance gets bigger.

Within the kriging routines it should be quite simple to implement this, since only the pairwise distances of the points are calculated. At this point we could install a switch to use the great-circle distance (or sth more fancy from geopy)

Also the variogram estimation should be easy adoptable, since only the distance calculation has to be adjusted.

Since no anisotropy and consequently no rotation is allowed for geographical coordinates, it would be a good idea to implement this as a switch in the CovModel. Then the following should be set:

- ``dim=2`` (here we could get a problem, since the above mentioned fix in RandMeth needs dim=3)
- ``anis=[1]``
- ``angles=[0]``

Maybe it could be a good idea to freeze them.

We could add a keyword-argument ``latlon`` or ``geo_coord`` to the ``CovModel.__init__``, that is ``False`` by default.

What do you think? @LSchueler, @fhesze, @AlrauneZ, @banesullivan

LSchueler:
Before putting too much work into this, we should actually check the differences between a naive approach like the one pasted below and the correct computation on a sphere in dependence of the area on which the fields are computed.

Does anyone have an idea what the largest areas are, on which random fields are generated? At least in oceanography, nobody cares about spherical coordinates for areas smaller than 50km x 50km.

Are you suggesting to neither support anisotropic nor rotated anisotropic fields? - That would reduce the potential applications dramatically.

This is just for visualisation, I would not suggest to generate a field for a quarter of the globe.

```python
import numpy as np
import plotly.graph_objects as go
from plotly.subplots import make_subplots

from gstools import SRF, Gaussian

def spherical_linspace(start, stop, steps):
    """Takes degrees, spits out rads."""
    return np.linspace(start * np.pi/180.0, stop * np.pi/180.0, steps)

def spherical_coords(theta, phi):
    x = R * np.cos(theta) * np.cos(phi)
    y = R * np.cos(theta) * np.sin(phi)
[...]

PR #98, opened 2020-07-27
Geo variogram

LSchueler:
A first step towards supporting spherical coordinate systems #54 is to support variogram estimation on spherical coordinate systems.

This is a first proposal for calculating unstructured variograms on spheres.

A few things to discuss:

- [x] is `spherical` a good name for the argument in the function `vario_estimate_unstructured`?
  - let's use `great-circle`
- [ ] how are we going to provide the new distance function to users? - This is important in order to, e.g. create bins.
  - Possible options:
    - import the Cython function from `estimator.pyx` into one of the `tools.py` namespaces
    - create a new `tools.pyx` as a more general Cython code collection
    - reimplement the distance function in Python
    - use on of the already existing Python packages as external dependenies, e.g. [haversine](https://github.com/mapado/haversine)
  - Suggestion:
    - [ ] users shouldn't need to care about the haversine routines (just use it internally)
    - [ ] the rest of GSTools should deal with xyz positions converted by a new routine `latlon2xyz`
    - [ ] variogram fitting needs to provide a new workflow to fit associated 3D *Yadrenko* models to estimated great-circle variograms
    - [ ] with all this, we don't need to rewrite the `SRF`, `CovModel` and `Krige` classes
    - [ ] we need a set of good tutorials dealing with geo-coordinates (explaining yadrenko models and great-circle distances)
- [ ] better testing: I'm not sure what else to test
- [ ] how could potential pitfalls / confusing arguments be caught (lat-lon, y-x)
  - as written above: conversions routines and good tutorials
- [ ] should we stick to the unit of metre or does kilometre make more sense?
  - we shoud use the great-circle distance in `deg` and leave this decision to the user

### DIFF
--- a/gstools/variogram/estimator.pyx
+++ b/gstools/variogram/estimator.pyx
@@ -10,7 +10,7 @@ import numpy as np
 cimport cython
 from cython.parallel import prange, parallel
 from libcpp.vector cimport vector
-from libc.math cimport fabs, sqrt
+from libc.math cimport fabs, sqrt, pow, sin, cos, atan2, M_PI
 cimport numpy as np
 
 
@@ -35,6 +35,27 @@ cdef inline double _distance_2d(
     const int j
 ) nogil:
     return sqrt((x[i] - x[j]) * (x[i] - x[j]) + (y[i] - y[j]) * (y[i] - y[j]))
+
+cdef inline double _distance_2d_haversine(
+    const double[:] x,
+    const double[:] y,
+    const double[:] z,
+    const int i,
+    const int j
+) nogil:
+    # x is latitude in degree
+    # y is longitude in degree
+    cdef double deg_2_rad = M_PI / 180.0
+    cdef double diff_lat = (x[j] - x[i]) * deg_2_rad
+    cdef double diff_lon = (y[j] - y[i]) * deg_2_rad
+    cdef double arg = (
+        pow(sin(diff_lat/2.0), 2) +
+        cos(x[i]*deg_2_rad) *
+        cos(x[j]*deg_2_rad) *
+        pow(sin(diff_lon/2.0), 2)
+    )
+    # earth radius for WGS84 ellipsoid r ~ 6371km
+    return 12742000.0 * atan2(sqrt(arg), sqrt(1.0-arg))
 
 cdef inline double _distance_3d(
     const double[:] x,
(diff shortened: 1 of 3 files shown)


gstools/variogram/estimator.pyx
```diff
@@ -36,6 +36,27 @@ cdef inline double _distance_2d(
...
+    cdef double diff_lat = (x[j] - x[i]) * deg_2_rad
+    cdef double diff_lon = (y[j] - y[i]) * deg_2_rad
+    cdef double arg = (
+        pow(sin(diff_lat/2.0), 2) +
+        cos(x[i]*deg_2_rad) *
+        cos(x[j]*deg_2_rad) *
+        pow(sin(diff_lon/2.0), 2)
+    )
+    # earth radius for WGS84 ellipsoid r ~ 6371km
+    return 12742000.0 * atan2(sqrt(arg), sqrt(1.0-arg))
```
MuellerSeb, line 58: I would use the great-circle distance in `deg` as the result, so we don't have to deal with the earth radius. This is also done in Pykrige this way ATM. Then we also don't have to care about `km` vs `m` and leave this decision to the user.

gstools/variogram/variogram.py
```diff
@@ -32,6 +32,18 @@ def _set_estimator(estimator):
         )
     return cython_estimator
 
+def _set_distance(distance):
+    """Translate the verbose Python distance identifier to single char."""
+    if distance.lower() == "euclidean":
+        cython_distance = "e"
+    elif distance.lower() == "spherical":
```
MuellerSeb, line 39: Same argument: Let's use `"great-circle"` here.

gstools/variogram/variogram.py
```diff
@@ -92,6 +105,13 @@ def vario_estimate_unstructured(
             * "cressie": an estimator more robust to outliers
 
         Default: "matheron"
+    distance : :class:`str`, optional
+        the distance function, possible choices:
+
+            * "euclidean": the distance in Euclidean spaces
+            * "spherical": the great-circle distance on spheres
```
MuellerSeb, line 112: Same argument: Let's use "great-circle" here.

gstools/variogram/variogram.py
LSchueler: Good idea

gstools/variogram/estimator.pyx
MuellerSeb: https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.haversine_distances.html

sklearn takes lat-lon in deg and returns to distance in rad. We could use that as a guide.

MuellerSeb:
this is now made obsolete by #113

PR #100, opened 2020-08-14
Isotropic rotation not possible

LSchueler:
I just wanted to make a pretty plot of a 3d field and instead of rotating the camera, I wanted to simply rotate the field.
This is not possible if the field is isotropic.

Of course, from a stochastic point of view, it doesn't make sense to rotate a field, if it is isotropic. But this is a restriction, which is not necessary, especially if one is working with a single realisation.

Especially for nice plotting, I suggest to simply remove the check for isotropicity, as seen in the diff.


LSchueler:
Thanks for the review!
The problem of converting the grid without rotation will hopefully be a thing of the past with the release of version 2.0!

PR #104, opened 2020-09-21
Variogram estimation from multiple fields

MuellerSeb:
With this PR a new option for unstructured variogram estimation is added:
You can now estimate one variogram from multiple fields on the same point-set. This could be useful to estimate a single variogram to describe multiple realizations of the same process.

Example:
- time series of precipitation data -> one variogram for all days needed

In addition, these fields can have missing values, denoted by a new `no_data` argument.

Checklist:
- [x] blackened
- [x] test written
- [x] examples added

Issue #83, opened 2020-04-22
vario_estimate_structured returns only nan if nan points in the field


TobiasGlaubach:
Hi,
not sure if this is intended behaviour, I was trying to exclude invalid points from a structured grid by setting them to `np.nan` before applying `vario_estimate_structured` the result was purely `np.nan`.

minimal code in python:

```python
import numpy as np
import gstools as gs
import matplotlib.pyplot as plt
from scipy import ndimage, misc

x = y = np.arange(100)
model = gs.Exponential(dim=2, var=1, len_scale=[12.0, 3.0], angles=np.pi / 8)
srf = gs.SRF(model, seed=20170519)
srf.structured([x, y])
val = srf.field
val[0,0] = np.nan
gamma_x = gs.vario_estimate_structured(val, direction="x")
print(gamma_x)

```

Thanks and cheers,
Tobi

LSchueler:
It's definitely not intended behaviour. Even if it was, it should be documented.
I'll see how to fix it. Thanks for reporting!

Here is the way I intended to cope with missing / invalid data:

``` python
import numpy as np
import gstools as gs
import matplotlib.pyplot as plt
from scipy import ndimage, misc

x = y = np.arange(100)
model = gs.Exponential(dim=2, var=1, len_scale=[12.0, 3.0], angles=np.pi / 8)
srf = gs.SRF(model, seed=20170519)
srf.structured([x, y])
val = srf.field
val_mask = np.zeros_like(val)
# mark invalid points
val_mask[0,0] = 1
val = np.ma.masked_array(val, mask=val_mask)
gamma_x = gs.vario_estimate_structured(val, direction="x")
print(gamma_x)
```

TobiasGlaubach:
Worked like a charm! Thanks for the workaround! 

While we are at it. Does GSTools have any functionality for estimating the rotation of a variogram? Currently I am rotating the grid before calcuation of the variogram.

```python
from scipy import ndimage
val_rot = ndimage.rotate(val, 71, reshape=True)
val_masked = np.ma.masked_array(val_rot, mask=val_rot == 0.0)
gs.vario_estimate_structured(val_masked , direction="x")
```
And then explore fits on the variograms calculated from those rotations.

LSchueler:
What exactly do you mean by "estimating the rotation of a variogram"?
[...]

PR #82, opened 2020-04-21
Make vario. estim. distance fct ready for N dim.

LSchueler:
As discussed in issue #77 and in the [PyKrige issue](https://github.com/GeoStat-Framework/PyKrige/issues/138) we want to support arbitrary field dimensions.

PR #106, opened 2020-11-07
Empirical Variogram: Update and Refactoring

MuellerSeb:
This PR reworks the whole variogram estimation subpackage:

- new routine name `vario_estimate` instead of `vario_estimate_unstructured` (old kept for legacy code) for simplicity
- new routine name `vario_estimate_axis` instead of `vario_estimate_structured` (old kept for legacy code) for simplicity
- **`vario_estimate`**
  - allow to pass multiple fields for joint variogram estimation (e.g. for daily precipitation) on same mesh
  - `no_data` option added to allow missing values
  - **masked fields**
    - user can now pass a masked array (or a list of masked arrays) to deselect data points.
    - in addition, a `mask` keyword was added to provide an external mask
  - **directional variograms**
    - diretional variograms can now be estimated
    - either provide a list of directions or angles for directions (spherical coordinates)
    - can be controlled by given angle tolerance and (optional) bandwidth
    - prepared for nD
  - an structured field (pos tuple describes axes) can now be passed to estimate an isotropic or directional variogram
  - distance calculation in cython routines in now independent of dimension
- **`vario_estimate_axis`**
  - estimation along array axis now possible in arbitrary dimensions
  - `no_data` option added to allow missing values (sovles #83)
  - axis can be given by name (`"x"`, `"y"`, `"z"`) or axis number (`0`, `1`, `2`, `3`, ...)

Thanks to @TobiasGlaubach for starting this (#87) and to @LSchueler for providing the first implementation of the new distance calculation in cython (#82). This also re-implements #104, which makes to original PR obsolete.


MuellerSeb:
@LSchueler one question: should a mask, that masks all values raise an ValueError, our should we return a 0 valued variogram and 0 counts for each bin? ATM it raises an ValueError.

gstools/variogram/estimator.pyx
```diff
@@ -102,100 +152,143 @@ cdef _normalization_func choose_estimator_normalization(str estimator_type):
...
+                dist = distance(dim, pos, j, k)
+                if dist >= bin_edges[i] and dist < bin_edges[i+1]:
+                    for d in range(d_max):
+                        if dir_test(dim, pos, dist, direction, angles_tol, bandwidth, k, j, d):
+                            for m in range(f_max):
+                                # skip no data values
+                                if not (isnan(f[m,k]) or isnan(f[m,j])):
+                                    counts[d, i] += 1
+                                    variogram[d, i] += estimator_func(f[m,k] - f[m,j])
+
```
MuellerSeb, line 214: One could argue to incorporate a `break` here, since if a point pair matches one direction, it (maybe) shouldn't be used in another direction as well. This only happens if the angels_tol is big enough, so two directions have an overlapping search area. @LSchueler your opinion?

LSchueler:
> @LSchueler one question: should a mask, that masks all values raise an ValueError, our should we return a 0 valued variogram and 0 counts for each bin? ATM it raises an ValueError.
> 
> EDIT: in order to make it consistent (if field is a masked array and the mask there is entirely True, a 0 valued variogram is returned), I removed the raised Error and return a 0 valued variogram without further calculations.

With what would this be consistent? - I'm not sure how meaningful a variogram of a completely invalid field is...

gstools/tools/geometric.py
```diff
@@ -161,3 +162,44 @@ def xyz2pos(x, y=None, z=None, dtype=None, max_dim=3):
...
+    if dim != angles.shape[1] + 1 or dim == 1:
+        raise ValueError("Wrong dim. ({}) for angles {}".format(dim, angles))
+    vec = np.empty((angles.shape[0], dim), dtype=dtype)
+    vec[:, 0] = np.prod(np.sin(angles), axis=1)
+    for i in range(1, dim):
+        vec[:, i] = np.prod(np.sin(angles[:, i:]), axis=1)  # empty prod = 1
+        vec[:, i] *= np.cos(angles[:, (i - 1)])
+    if dim in [2, 3]:
+        vec[:, [0, 1]] = vec[:, [1, 0]]  # to match convention in 2D and 3D
+    return vec
```
LSchueler, line 205: I hope I will never have to fix a bug in here ;-)

gstools/variogram/estimator.pyx
LSchueler: Every hardcore Bayesian would probably cry "heresy!", but maybe it is exactly the intention of someone using a large tolerance to do a preliminary test without any knowledge about the main directions of the field. Then, I guess it would be helpful to use as many data points as possible, even if they overlap and are redundant.

LSchueler (changes requested):
Once again a tremendous effort! Thanks for that.
I think I like the renaming of the estimation functions, but it could be that users get even more confused about the struct, unstruct stuff, let's see.

I hope you don't mind that I pushed some small typo-fixes directly onto this branch.

I think if you update the changelog, I'd be happy about a merge.

gstools/tools/geometric.py
MuellerSeb: As I said: rotation in higher dimensions (>1) is a pain in the neck!

gstools/variogram/estimator.pyx
MuellerSeb: For speed up reasons we could check all angles between directions and if their maximum is less than  `2 * angels_tol`, we could break ;-)

MuellerSeb:
> > @LSchueler one question: should a mask, that masks all values raise an ValueError, our should we return a 0 valued variogram and 0 counts for each bin? ATM it raises an ValueError.
> > EDIT: in order to make it consistent (if field is a masked array and the mask there is entirely True, a 0 valued variogram is returned), I removed the raised Error and return a 0 valued variogram without further calculations.
> 
> With what would this be consistent? - I'm not sure how meaningful a variogram of a completely invalid field is...

If the field is a masked array and the mask there is entirely True, a 0 valued variogram is returned. If an external mask was given that is enterly true with an ordinary field, an Error was raised.

That was the inconsistency. The now returned variogram is constantly 0 and the counts are 0 as well. I think this is meaningful, since it says: No data!

MuellerSeb:
> Once again a tremendous effort! Thanks for that.

Thanks :heart: 

> I think I like the renaming of the estimation functions, but it could be that users get even more confused about the struct, unstruct stuff, let's see.

I think it is now in line with all other routines, since the `mesh_type` arguments was added and you can feed struct and unstruct fields to `vario_estimate`. The `vario_estimate_axis` was originally not for structured fields (rectilinear grids in our case) but for "structured points" (in the sence of VTK), where every axis is equidistant. So I think with the `axis` hint (like axis of numpy arrays), the intention of this routine is clearer now.

> I hope you don't mind that I pushed some small typo-fixes directly onto this branch.

Of course not. My english is not se goodest.

> I think if you update the changelog, I'd be happy about a merge.

I'll do a changelog later in develop, so I can mention the PRs.

LSchueler (approved):

gstools/variogram/estimator.pyx
MuellerSeb: Fixed that by checking if the directions are separated. If the search bands overlap, point pairs are counted repeatedly.

PR #107, opened 2020-11-10
CI update: provide Python 3.9 wheels

MuellerSeb:
This PR introduces wheels for Python 3.9.

In addition, pushes to `test.pypi.org` are now only done for pushes to `master` or `develop`.

Also a `FutureWarning` from numpy for element wise comparison of arrays is fixed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,7 +26,7 @@ before_install:
     fi
 
 script:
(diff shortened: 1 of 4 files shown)


LSchueler (approved):
Go for it!

Issue #90, opened 2020-05-09
[Refactor] prefere "cor" to specify userdefined CovModel


MuellerSeb:
At the moment we allow to specify one of the following functions for a user-defined model:
- `variogram` - full variogram including variance, nugget, length-scale and potential additional parameters 
- `covariance` - covariance only including variance, length-scale and potential additional parameters
- `correlation` - correlation only including length-scale and potential additional parameters
- `cor` - normalized correlation depending on the non-dimensional distance only including potential additional parameters

Only the `cor` method is a save way of defining a fully working CovModel which results in the standard definition from the [documentation](https://geostat-framework.readthedocs.io/projects/gstools/en/stable/examples/02_cov_model/index.html#tutorial-2-the-covariance-model), so we should think about sticking to this one as the standard way of defining a CovModel.

mmaelicke:
Could you elaborate on why only `cor` is a save way? I'm not into `CovModel` that much. 

In Terms of GS-Framework v2 this might be helpful to limit the user-defined models of `CovModel` to `cor`. A user-defined `variogram` can then be used to define a potential `gstools.Variogram` instance. Would make it way easier to map between a `skgstat.Variogram` and `gstools.Variogram`, or merging the two at that point, though. (should go into another issue)

There are also instances, where a covariance function is not defined, but a variogram is. From my point of view, it would also be a more clearer notation if not *any* `variogram` can be used to define a `CovModel`.  In these cases any kind of `Variogram` class can be used to decide if a `CovModel` would be defined or not.

MuellerSeb:
The only valid variograms models, that we are excluding, are the unbounded ones. So if we want to support unbounded variograms, I think this should be represented by a new class. 

The RandMeth class only takes bounded variograms to generate randomfields, since it is depending on the spectral density of the covariance. And the covariance only exists for bounded variograms. In addition, the spectral density is only a real "density" function, if the covariance/varigram model is valid (conditional negative semidefinite).

If we implement unbounded variograms, we should also provide a SRF generator for these. (Sequential Gaussian, Turning bands, ... something like this)

MuellerSeb:
@mmaelicke : The variogram in GSTools is represented by:
```
vario(r) = var * (1 - cor(r / len_scale)) * nugget
```

If you use the CovModel class to create your own variogram, and you override the variogram method to define it, you have to take care of variance, len_scale and nugget, so everything is still working (fitting, random field generation, integral_scale resetting and so on...)

[...]

Issue #88, opened 2020-05-03
[Enhancement] more covariance models


MuellerSeb:
There are far more covariance models described in literature, than provided by GSTools.
I would suggest adding the following:
- [x] `JBessel` class (hole effect model valid in arbitrary dimensions)
- [x] simple truncated powerlaw `TPL`, given by `(1-h)^nu` valid for `nu >= (d+1)/2` (extension to truncated linear)
- [x] `HyperSpherical` (replacement of the current `Intersection` model)
- [x] `SuperSpherical` (HyperSpherical with freely selectable parameter)

More ideas?

Issue #86, opened 2020-05-02
[Enhancement] dimensionality check for CovModel


MuellerSeb:
Since some CovModels are only valid up to a certain dimensionality (eg. Linear only in 1D, Circular only in nD for n<3), we should provide a `dim_valid` routine in the `CovModel` class, that could also be overridden by the user.

As default, we should return `True` since many models are valid in arbitrary dimensions.

MuellerSeb:
A warning should pop up. An Error would be to harsh, since users should be allowed to make their own mistakes.

PR #109, opened 2020-11-11
CovModel: Update and Refactoring

MuellerSeb:
This PR updates and refactors the covariance models:
- Fitting routines now support input of directional **empirical variograms** to estimate anisotropy
- add new `rescale` argument and attribute to the `CovModel` class to be able to rescale the `len_scale` (usefull for unit conversion or rescaling `len_scale` to coincide with the `integral_scale` like it's the case with the Gaussian model)
  See: #90, https://github.com/GeoStat-Framework/PyKrige/issues/119
- added new `len_rescaled` attribute to the `CovModel` class, which is the rescaled `len_scale`: `len_rescaled = len_scale / rescale`
- new method `default_rescale` to provide default rescale factor (can be overridden)
- remove `doctest` calls
- docstring updates in CovModel and derived models
- updated all models to use the `cor` routine and make use of the `rescale` argument (See: #90)
- TPL models got a separate base class to not repeat code
- added **new models** (See: #88):
  -  `HyperSpherical`: (Replaces the old `Intersection` model) Derived from the intersection of hyper-spheres in arbitrary dimensions. Coincides with the linear model in 1D, the circular model in 2D and the classical spherical model in 3D
  - `SuperSpherical`: like the HyperSpherical, but the shape parameter derived from dimension can be set by the user. Coincides with the HyperSpherical model by default
  - `JBessel` a hole model valid in all dimensions. The shape parameter controls the dimension it was derived from. For `nu=0.5` this model coincides with the well known `wave` hole model. 
  - `TPLSimple` a simple truncated power law controlled by a shape parameter `nu`. Coincides with the truncated linear model for `nu=1`
- string representation of the `CovModel` class is new using a float precision (`CovModel._prec=3`) to truncate longish output
- dimension validity check: raise a warning, if given model is not valid in the desired dimension (See: #86) 
- more comprehensive testing


MuellerSeb:
@LSchueler are you willing to take a first look?

LSchueler (changes requested):
This adds a lot of really nice little things to the project!
I think the len_rescale is a very elegant solution to quite some minor problems. - Cool!
But I think you should add some tests for the interactions of the arguments `len_scale`, `integral_scale`, and `len_rescale`.

I fixed some typos directly on this branch... as usual ;-)

MuellerSeb:
@LSchueler thanks :wink:
So I will add a test for `rescale`. Fitting should be tested better as well.

MuellerSeb:
Coverage at 92% now! :tada:

PR #112, opened 2020-11-18
[Field] refactoring and update

MuellerSeb:
# Overall
- drop usage of `pos2xyz` and `xyz2pos`
# Generators
- remove structured option from generators (structured pos need to be converted first)
- explicitly assert dim=2,3 when generating vector fields
# Tools
- remove routines `pos2xyz` and `xyz2pos` (finally)
- add routines to format struct. pos tuple by given `dim` or `shape`
- add routine to format struct. pos tuple by given `shape` (variogram helper)
- remove `field.tools` subpackage
# Field
- simplify `pre_pos` routine to save pos tuple and reformat it an unstructured tuple
- simplify field shaping
- simplify plotting routines
# SRF/Krige
- simplify generation and conditioning
# Mesh
- support `meshio>=4.0`
- add meshio as dependency
- add example


MuellerSeb:
@LSchueler should be ready for a first glimpse.

MuellerSeb:
@LSchueler there is still a bug in `format_unstruct_pos_shape` that needs to be tackled by me. Just to keep that in mind.

examples/01_random_field/06_higher_dimensions.py
```diff
@@ -0,0 +1,75 @@
+"""
+Higher Dimensions
+-----------------
+
+GSTools provides experimental support for higher dimensions.
```
LSchueler, line 5: That sounds fancy! :-)

gstools/field/generator.py
```diff
@@ -441,62 +406,3 @@ def _create_unit_vector(self, broadcast_shape, axis=0):
         e1 = np.zeros(shape)
         e1[axis] = 1.0
         return e1
-
```
LSchueler, line 444: Sooo nice!

gstools/krige/tools.py
```diff
@@ -44,17 +41,19 @@ def no_trend(*args, **kwargs):
     return 0.0
 
 
-def eval_func(func, pos, mesh_type="structured"):
+def eval_func(func, pos, dim, mesh_type="structured"):
```
LSchueler, line 44: Would it make sense to put the argument `dim` to the first position to make it consistent with e.g. `CovModel`?

gstools/krige/tools.py
```diff
@@ -63,16 +62,16 @@ def eval_func(func, pos, mesh_type="structured"):
...
+        pos, shape = format_struct_pos_dim(pos, dim)
+        pos = gen_mesh(pos)
+    else:
+        pos = np.array(pos, dtype=np.double).reshape(dim, -1)
+        shape = np.shape(pos[0])
+    return np.reshape(func(*pos), shape)
 
 
-def set_condition(cond_pos, cond_val, max_dim=3):
+def set_condition(cond_pos, cond_val, dim):
```
LSchueler, line 74: Would it make sense to put the argument `dim` to the first position to make it consistent with e.g. `CovModel`?

tests/test_randmeth.py
```diff
@@ -30,64 +30,64 @@ def setUp(self):
...
+
+    # def test_struct_3d(self):
+    #     modes = self.rm_3d(
+    #         self.x_grid, self.y_grid, self.z_grid, mesh_type="structured"
+    #     )
+    #     self.assertAlmostEqual(modes[0, 0, 0], 0.55488481)
+    #     self.assertAlmostEqual(modes[0, 1, 0], 0.41858766)
+    #     self.assertAlmostEqual(modes[1, 1, 0], 0.95133855)
+    #     self.assertAlmostEqual(modes[0, 1, 1], 0.65475042)
+    #     self.assertAlmostEqual(modes[1, 1, 1], 1.40915120)
```
LSchueler, line 67: Are you planing on putting similar tests to `test_srf.py`?

tests/test_incomprrandmeth.py
```diff
@@ -31,47 +31,47 @@ def setUp(self):
...
+    #     unstr_modes = self.rm_2d(x_tuple, y_tuple, mesh_type="unstructured")
+    #     str_modes = self.rm_2d(x_grid, y_grid, mesh_type="structured")
+    #     for d in range(2):
+    #         k = 0
+    #         for i in range(len(x_grid)):
+    #             for j in range(len(y_grid)):
+    #                 self.assertAlmostEqual(
+    #                     str_modes[d, i, j], unstr_modes[d, k]
+    #                 )
+    #                 k += 1
```
LSchueler, line 74: Are you planing on putting similar tests to `test_srf.py`?

LSchueler (changes requested):
That's already looking so great!

As always, I've fixed some typos, so remember to pull before doing new commits.

I guess the `gstools.field.summator.summate_{unstruct,summate_incompr_unstruct}` should be renamed. The `_unstruct` is probably only confusing from now on. Or what reasons are there to keep the name? The name could be `summate_modes` if you don't like `summator.summate`.

Apart from that, I don't have too much to add at the moment, except maybe: great job!

tests/test_randmeth.py
MuellerSeb: Since this (low-level) functionality is not present anymore, we don't need to test it anymore.

tests/test_incomprrandmeth.py
MuellerSeb: Since this (low-level) functionality is not present anymore, we don't need to test it anymore.

gstools/krige/tools.py
MuellerSeb: I can do so, but this is a low-level function users shouldn't use. But following [Hyrum's Law](https://github.com/dwmkerr/hacker-laws#hyrums-law-the-law-of-implicit-interfaces) it could be a good idea to be consistent.

gstools/krige/tools.py
MuellerSeb: I can do so, but this is a low-level function users shouldn't use. But following [Hyrum's Law](https://github.com/dwmkerr/hacker-laws#hyrums-law-the-law-of-implicit-interfaces) it could be a good idea to be consistent.

gstools/field/base.py
```diff
@@ -73,8 +68,8 @@ def unstructured(self, *args, **kwargs):
         return call(*args, **kwargs)
 
     def mesh(
-        self, mesh, points="centroids", direction="xyz", name="field", **kwargs
-    ):  # pragma: no cover
+        self, mesh, points="centroids", direction="all", name="field", **kwargs
```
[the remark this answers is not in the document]
MuellerSeb: To be backward compatible xyz are the only alternative labels for dimensions ;-)

gstools/__init__.py
```diff
@@ -104,7 +104,7 @@
    vario_estimate
    vario_estimate_axis
 """
-
+# Hooray!
```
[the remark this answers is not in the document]
MuellerSeb: Just kidding: https://github.com/GeoStat-Framework/GSTools/commit/7523ae0b10cff67d54c3d6ea0de04738f60a9921

examples/01_random_field/06_higher_dimensions.py
MuellerSeb: It is!

gstools/krige/tools.py
MuellerSeb: Thinking about it twice, I think we should keep it this way. It is low-level and the first parameters should hold the conditions, as the function name states.

gstools/krige/tools.py
MuellerSeb: Thinking about it twice, I think we should keep it this way. It is low-level and the first parameter should hold the function to evaluate, as the function name states.

PR #114, opened 2020-11-30
Migrate from TravisCI to GitHub Actions

LSchueler:
Unfortunately, TravisCI wants to charge us for their CI tools. Therefore, I've migrated to GitHub Actions.

Nearly everything is the same as before, except the parallel wheel building by changing the environment variable `CIBW_BUILD`. That is a real pain in the neck in Actions at the moment, therefore the wheel building is done sequentially. But this can be changed in the future.
A build takes about 20 min. at the moment.

For the actions results have a look in the fork:

https://github.com/LSchueler/GSTools/actions


requirements_setup.txt
```diff
@@ -1,4 +1,5 @@
 setuptools>=41.0.1
 setuptools_scm>=3.5.0
+cibuildwheel>= 1.7.0
```
MuellerSeb: This is actually not an setup requirement. Maybe as CI requirement? :wink:

requirements_test.txt
```diff
@@ -1,2 +1,3 @@
 pytest-cov>=2.8.0
 pytest>=5.3.0
+coveralls>=2.0.0
```
MuellerSeb, line 3: also only a CI req?

.github/workflows/main.yml
```diff
@@ -0,0 +1,116 @@
+name: Continuous Integration
+
+on: [push, pull_request]
+
+env:
+  # needed by coveralls
+  GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+  CIBW_BUILD: "cp36-* cp37-* cp38-*"
```
MuellerSeb, line 8: No py35? No py39? Guess they already dropped py35, didn't they?

.github/workflows/main.yml
```diff
@@ -0,0 +1,116 @@
...
+          pip install cibuildwheel>=1.7.0
+      - name: Build wheels
+        run: |
+          python -m cibuildwheel --output-dir dist
+
+      - uses: actions/upload-artifact@v2
+        with:
+          path: ./dist/*.whl
+
+  build_sdist:
```
MuellerSeb, line 47: Why running sdist on all os and py versions? I think py38 on linux should be enough.

.github/workflows/main.yml
```diff
@@ -0,0 +1,116 @@
...
+
+    steps:
+      - uses: actions/download-artifact@v2
+        with:
+          name: artifact
+          path: dist
+
+      - name: Publish to Test PyPI
+        # only if working on develop
+        # TODO change from master to develop
```
MuellerSeb, line 101: Can you resolve this TODO?

MuellerSeb (changes requested):
This looks really good! Thank you :heart: I was so in fear of this migration.
Still found some things to change.

.github/workflows/main.yml
LSchueler: Ups... that got lost during a rebase session I guess, thank!

.github/workflows/main.yml
LSchueler: Testing OpenMP on all combinations was my intention...

.github/workflows/main.yml
LSchueler: Yeah, we'd have to use an older CIBuildWheel version. I know we had a discussion about keeping py35. You think we should stick to it?

requirements_test.txt
LSchueler: Changed

.github/workflows/main.yml
MuellerSeb: Oh... clever ;-)

.github/workflows/main.yml
MuellerSeb: I think they still **build** wheels for py35, they just don't run CIBuildWheel under py35 AFAICT from their doc.

.github/workflows/main.yml
MuellerSeb: But then: why not py39?

.github/workflows/main.yml
LSchueler: Added both, py35 and py39

.github/workflows/main.yml
LSchueler: ... a little bit embarrassing...I missed the release of py39... :blush:
I'll let the current actions jobs run through and then add py39

LSchueler:
Everything seems to be working now. Nothing was published on Test PyPI, because I'm working on the master branch. That will change after the merge.

MuellerSeb (approved):
Nice! :fireworks:
You could merge this now! :wink:

PR #123, opened 2020-12-17
Add new example for generating synthetic P field

LSchueler:
I've created a new example showing how to generate synthetic precipitation fields with GSTools.

I'm not totally sure of where to put the example. I thought it would be too complex to put it into the transformation part. Maybe it's nice to keep it as a little bit more complex field generation example.

But I also think it's too simple to put it into the GeoStat-Examples repo.


examples/01_random_field/07_precipitation.py
```diff
@@ -0,0 +1,114 @@
...
+series.
+"""
+
+import copy
+import numpy as np
+import matplotlib.pyplot as plt
+import gstools as gs
+
+# fix the seed for reproducibility
+seed = 20170521
```
MuellerSeb, line 21: What movie? :wink:

examples/01_random_field/07_precipitation.py
```diff
@@ -0,0 +1,114 @@
...
+###############################################################################
+# Finally we can have a look at the fields resulting from each step. For a
+# closer look, we will examine a cross section at an arbitrary location. And
+# afterwards we will create a contour plot for visual candy.
+
+fig, axs = plt.subplots(2, 2, sharex=True, sharey=True)
+
+axs[0,0].set_title('Gaussian')
+axs[0,0].plot(t, P_gau[:, 20])
+axs[0,0].set_ylabel(r'$P$ / mm')
```
MuellerSeb, line 72: I was a bit confused by the `/` as separation for the unit. What do you think about `P [mm]`?

examples/01_random_field/07_precipitation.py
```diff
@@ -0,0 +1,114 @@
...
+cbar.ax.set_ylabel(r'$P$ / mm')
+axs[1,0].set_xlabel(r'$t$ / d')
+axs[1,0].set_ylabel(r'$x$ / km')
+
+fig.delaxes(axs[1,1])
+plt.tight_layout()
+
+###############################################################################
+# In this example we have created precipitation fields which have one spatial
+# dimension, but it is very easy to add a second spatial dimension.
```
MuellerSeb, line 114: There already came mails asking about how to do stuff in 2D where there was only a 1D example.. maybe we should skip this sentence or provide a minimal example for a field in 2D. What do you think?

MuellerSeb (changes requested):
This is a really nice example! We need more of these :wink:
What do you think about putting it in `misc` ?
But I am also fine with the way it is!

examples/01_random_field/07_precipitation.py
LSchueler: Of course Lynch's newest masterpiece Twin Peaks the Return

examples/01_random_field/07_precipitation.py
LSchueler: It's the best and mathematically most rigorous way of labeling axes! Let me explain:
E.g., you read from the axes and the graph at a certain point in time: _P_ / mm = 40 You divide the equation by the unit mm and get
_P_ = 40mm
tada!

But yeah, I had to already convince other people, like coauthors of this way of life, as I like to call it ;-)

examples/01_random_field/07_precipitation.py
LSchueler: You are probably right. All add some lines. Hmm, maybe I can use pyvista for visualization?

examples/01_random_field/07_precipitation.py
MuellerSeb: Sneaky! Ok, you got me. Like it! Will do so too from now on! :+1:

examples/01_random_field/07_precipitation.py
MuellerSeb: Would use a simple 2D plot with a masked array to get something like a weather forecast plot for rain for a single time-step :wink:
Or an animation? :thinking: sphinx-gallery has support for matplotlib animations: https://sphinx-gallery.github.io/stable/auto_examples/plot_8_animations.html

LSchueler: Yeah, it would be pretty cool, to do something like that. I just took the easiest path with the current commits, but I'm not convinced.
I'll think about a better visualisation.

LSchueler:
Okay, I moved the tutorial to "misc", I totally forgot about that folder...
I also made a suggestion with some renamed examples and left out the numbering of the tutorials. I don't really see how that helps. They don't really build up on each other.
What do you think?

What do you think about moving the Herten example to misc too? - It's in the variogram estimation folder at the moment, but it's more of a complete workflow example.

examples/01_random_field/07_precipitation.py
MuellerSeb: It is also nice, that this is implicitly the first example for spatio-temporal fields with a joint model! I guess we should start here for supporting spatio-temporal tools:
https://www.rdocumentation.org/packages/gstat/versions/2.0-6/topics/vgmST

MuellerSeb:
> Okay, I moved the tutorial to "misc", I totally forgot about that folder...
> I also made a suggestion with some renamed examples and left out the numbering of the tutorials. I don't really see how that helps. They don't really build up on each other.
> What do you think?

A bit of succession is given, but you are right, we don't need to enumerate these chapters but their order is still preserved.

> What do you think about moving the Herten example to misc too? - It's in the variogram estimation folder at the moment, but it's more of a complete workflow example.

Should do so. There will be a more complex Herten example with the GSTools paper either way.

LSchueler:
I reshuffled the examples. What do you say, are we ready to merge now?

MuellerSeb:
LGTM! :+1:

MuellerSeb (approved):
Go!

MuellerSeb:
@LSchueler both of "let's get this merged" and "looks good to me" ... I am totally into this internet slang :smile:

PR #125, opened 2021-01-05
CI: The sdists should only be uploaded once to the artifact

LSchueler:
The `build_sdist` job in the actions script uploads the artifact for every Python version. Often this works without problems, but sometimes it comes to errors and the artifact is not created.
The reason why I keep the sdist creation with all Python versions is to test GSTools without OpenMP support.

This PR is a draft, but the only way to kick off the actions.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -49,7 +49,6 @@ jobs:
           python -m cibuildwheel --output-dir dist
 
       - uses: actions/upload-artifact@v2
-        if: ${{ matrix.python-version }} == '3.9'
         with:
           path: ./dist/*.whl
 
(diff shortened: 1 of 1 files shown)


MuellerSeb:
I guess we need to use something like `if ${{ matrix.python-version }} == 3.9 then <sdist>...`

.github/workflows/main.yml
```diff
@@ -49,6 +49,7 @@ jobs:
           python -m cibuildwheel --output-dir dist
 
       - uses: actions/upload-artifact@v2
+        if: ${{ matrix.python-version }} == '3.9'
```
MuellerSeb, line 52: Thats the wrong place.

.github/workflows/main.yml
```diff
@@ -91,6 +92,7 @@ jobs:
 
       - uses: actions/upload-artifact@v2
         with:
+          python-version: 3.9
```
MuellerSeb, line 95: This doesn't work. Try `if: ${{ matrix.python-version }} == '3.9'` here (above with)

.github/workflows/main.yml
LSchueler: :-) Yeah thanks... somehow my text editor must have mysteriously scrolled up...

LSchueler:
Should be working now. Do you approve?

MuellerSeb (approved):
Yeah! That does the trick. :+1:

Issue #128, opened 2021-01-12
[RandMeth] switch to select sampling strategy


MuellerSeb:
Following the discussion in #126, we should implement a switch to the `RandMeth` class, to select the sampling strategy.

Proposal: Introducing an argument `sampling` to the `RandMeth.__init__` signature with the following options:
- `"auto"` (default): current behavior (if ppf is defined use inversion method, otherwise mcmc)
- `"inversion"`: use inversion method (even if only pdf is defined)
- `"mcmc"`: use mcmc (even if ppf is defined)

@LSchueler fine?

Issue #126, opened 2021-01-11
Systematic artifacts with Exponential field generation


leonfoks:
I'm trying to generate a 2D field using the Exponential generator, however there seem to be artifacts when compared to a Gaussian.

A minimal example that works and will generate a smooth image.
```Python
import matplotlib.pyplot as plt
import gstools as gs
import numpy as np 
x = range(1024)
y = range(128)
model = gs.Gaussian(dim=2, var=1.0, len_scale=[1000,8], angles=-np.pi/100)
srf = gs.SRF(model, seed=201705320, upscaling='coarse_graining')
srf.structured([x, y])
srf.plot()
plt.axis()
```
![Figure_0](https://user-images.githubusercontent.com/19690642/104329678-744b3400-54ed-11eb-88a4-de1ca708e687.png)

If I simply switch to an Exponential, there appears to be a systematic artifact that is generated in the image.
```Python
import matplotlib.pyplot as plt
import gstools as gs
import numpy as np 
x = range(1024)
y = range(128)
model = gs.Exponential(dim=2, var=1.0, len_scale=[1000,8], angles=-np.pi/100)
srf = gs.SRF(model, seed=201705320, upscaling='coarse_graining')
srf.structured([x, y])
srf.plot()
plt.axis()
```
![Figure_1](https://user-images.githubusercontent.com/19690642/104329701-7f05c900-54ed-11eb-8796-3a513c5ae958.png)

I have played with a few things to rule out the call arguments, e.g. same discretization (field.shape), equal length len_scale, angle=None, and plotting with axis('equal').  There is always what appears to be a systematic (deterministic?) undulation.

Am I correct here? Or am I missing something?  My goal is to create a wider than tall cross section of something that resembles horizontal stratigraphy.  The discretization is different here by design.

Thanks!

MuellerSeb:
Hey @leonfoks!

My educated guess is, that we see the modes of the underlying Randomization Method.
The fields are created by a superposition of sin and cos waves. To few of them can create artifacts like the one you encountered.
You can change that number of modes when initializing the SRF class:

```python
import gstools as gs
import numpy as np

x = range(1024)
y = range(128)
model = gs.Exponential(dim=2, var=1.0, len_scale=[1000,8], angles=-np.pi/100)
srf = gs.SRF(model, seed=201705320, upscaling='coarse_graining', mode_no=10000)
field = srf.structured([x, y])
ax = srf.plot()
ax.set_aspect("equal")
```
[...]

PR #130, opened 2021-01-13
New Class for Conditioned Spatial Random Fields

MuellerSeb:
Closes: #128, #126 

This PR adds a new class for conditioned random fields `CondSRF`.

This class replaces the conditioning feature of the SRF class, which was limited to Ordinary and Simple kriging.

`CondSRF` behaves similar to the `SRF` class, but instead of a covariance model, it takes a kriging class as input. With this kriging class, all conditioning related settings are defined.

During refactoring the `RandMeth` class, a switch to select the sampling strategy was added. 

# Note
This change breaks backward compatibility, since conditioned random fields are no longer provided by the `SRF` class.


MuellerSeb:
@LSchueler this is ready for review. :relaxed:

gstools/field/cond_srf.py
```diff
@@ -0,0 +1,170 @@
...
+import numpy as np
+from gstools.field.generator import RandMeth
+from gstools.field.base import Field
+from gstools.krige import Krige
+
+__all__ = ["CondSRF"]
+
+GENERATOR = {
+    "RandMeth": RandMeth,
+}
```
LSchueler, line 23: Maybe add a comment stating that this is the dict. of supported generators?

gstools/field/cond_srf.py
```diff
@@ -0,0 +1,170 @@
...
+        Name of the field generator to be used.
+        At the moment, the following generators are provided:
+
+            * "RandMeth" : The Randomization Method.
+              See: :any:`RandMeth`
+            * "IncomprRandMeth" : The incompressible Randomization Method.
+              This is the original algorithm proposed by Kraichnan 1970
+              See: :any:`IncomprRandMeth`
+            * "VectorField" : an alias for "IncomprRandMeth"
+            * "VelocityField" : an alias for "IncomprRandMeth"
```
LSchueler, line 43: These should not be listed here.

gstools/krige/base.py
```diff
@@ -16,7 +16,7 @@
 from scipy.spatial.distance import cdist
 import scipy.linalg as spl
 from gstools.field.base import Field
-from gstools.krige.krigesum import krigesum
+from gstools.krige.krigesum import krigesum, krigesumfast
```
LSchueler, line 19: `krigesum_fast`?

gstools/krige/krigesum.pyx
```diff
@@ -37,3 +37,28 @@ def krigesum(
             field[k] += cond[i] * krig_fac
 
     return np.asarray(field), np.asarray(error)
+
+
+def krigesumfast(
```
LSchueler, line 42: I think a short comment on what the difference to `krigesum` is should be written here.
Have you measured the difference? - Maybe it's more `krigesumwithlessmemoryconsumption`? ;-)

gstools/random/rng.py
```diff
@@ -61,10 +61,10 @@ def sample_ln_pdf(
         nwalkers : :class:`int`, optional
             The number of walkers in the mcmc sampler. Used for the
             emcee.EnsembleSampler class.
-            Default: 100
+            Default: 50
         burn_in : :class:`int`, optional
             Number of burn-in runs in the mcmc algorithm.
-            Default: 100
+            Default: 20
         oversampling_factor : :class:`int`, optional
```
LSchueler, line 68: It's funny how some topics overlap ;-)

gstools/random/tools.py
```diff
@@ -46,13 +46,9 @@ def seed(self):
...
         return self._seed
 
-    def __str__(self):
-        """Return String representation."""
-        return self.__repr__()
-
     def __repr__(self):
         """Return String representation."""
-        return "RNG(seed={})".format(self.seed)
+        return "MasterRNG(seed={})".format(self.seed)
```
LSchueler, line 51: Good catch!

LSchueler (changes requested):
Wow, the `SRF` class is looking soo much nicer now!

Really nice work!

I suggest to import `gstools.field.base.Field` to `gstools.field`. The import in one of the examples looks a but ugly.

I just made a few small remarks in the code. I think after you have addressed we are ready to go!!

gstools/random/rng.py
MuellerSeb: I couldn't resist...

gstools/krige/krigesum.pyx
MuellerSeb: I am also not happy. Got a better name/idea?

gstools/field/cond_srf.py
MuellerSeb: Good catch!

gstools/krige/krigesum.pyx
LSchueler: Hmm, what about something in the direction of `calc_krige_and_error_field` and `calc_krige_field`?

I was just thinking, what about combining both functions with a bool `error` or `error_field` or `calc_error_field` as an argument? - I mean, with predictive branching, it shouldn't be much slower and with `error = False`, the array could be set to `cdef double[:] error = np.zeros(1)` or something for saving memory.
I guess that would have to be profiled to see if it actually is fast.
But that's also something very low on the priority list.

examples/06_conditioned_fields/00_condition_ensemble.py
```diff
@@ -16,30 +16,35 @@
...
+#
+# Here we use a gaussian covariance model and ordinary kriging for conditioning
+# the spatial random field.
 
-# spatial random field class
 model = gs.Gaussian(dim=1, var=0.5, len_scale=1.5)
-srf = gs.SRF(model)
-srf.set_condition(cond_pos, cond_val, "ordinary")
+krige = gs.krige.Ordinary(model, cond_pos, cond_val)
+csrf = gs.CondSRF(krige)
```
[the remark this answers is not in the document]
MuellerSeb: Done.

examples/06_conditioned_fields/00_condition_ensemble.py
```diff
@@ -16,30 +16,35 @@
...
 
 ###############################################################################
 
 fields = []
 for i in range(100):
     # print(i) if i % 10 == 0 else None
```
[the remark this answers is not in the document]
MuellerSeb: Done.

gstools/field/cond_srf.py
MuellerSeb: Done.

gstools/field/cond_srf.py
MuellerSeb: Done.

gstools/krige/base.py
MuellerSeb: Renamed to `calc_field_krige_and_variance` and `calc_field_krige`.

gstools/krige/krigesum.pyx
MuellerSeb: Renamed to `calc_field_krige_and_variance` and `calc_field_krige`. Added method `_summate` to `Krige` to care about selection of the correct routine.

LSchueler (approved):
:-) Found another typo ;-)

But anyway, here is a big thumbs up!

Issue #55, opened 2020-01-10
Automatic binning


MuellerSeb:
This is just a note, so we don't forget and can talk about it:

My proposal for default bining:

- number of bins calculated by either:
  - (a) Sturges rule:

        n_bins = 2 * np.log2(n_points) + 1

  - (b) Rice rule:

        n_bins = np.power(n_points, 2 / 3)

- maximal bin edge as a third of the box diameter:

      diam = np.sqrt((x.max - x.min)**2 + (y.max - y.min)**2 + (z.max - z.min)**2)
      bin_max = diam / 3

- resulting bin sizes (quadratic growth):

      bins = np.linspace(0, bin_max**(1/2), n_bins)**2

Example: 30 points with ``diam = 900`` and Struges rule:

    bins = np.linspace(0, (900/3)**0.5, int(np.around(2*np.log2(30) + 1))) ** 2
    array([  0.,   3.,  12.,  27.,  48.,  75., 108., 147., 192., 243., 300.])

banesullivan:
@MuellerSeb, what is the `d` in 

```
diam = ((x.max - x.min)**d + (y.max - y.min)**d + (z.max - z.min)**d)**(1/d)
```

MuellerSeb:
The dimension... sorry

MuellerSeb:
It should be contantly 2, since it is the euclidean norm.. I just mixed it up with the d-norm. 

Comment was updated.

banesullivan:
@MuellerSeb, I did this for the example on #36 and it has really great results. The variogram doesn't look all that pretty, but the results were compelling.

I will make a PR with this now

MuellerSeb:
Thanks for your PR. @LSchueler is working on this at the moment, so I have closed your PR in favor of the re-written routines.

But it is good to hear, that the scheme was working for you! That means we are on the right way.

MuellerSeb:
@LSchueler: It would be nice to get some information on the variance within each bin during the variogram estimation. One could use a sequential update during the estimation, so we don't consume memory:

![mean](https://wikimedia.org/api/rest_v1/media/math/render/svg/72192362c590b89684309e78f6644ab0864f8d1b)
![var](https://wikimedia.org/api/rest_v1/media/math/render/svg/a07187029f774ee0da058f3b6ab6e5d3e7131df4)

We could later use this variance as a target function to estimate anisotropy and rotation by minimizing it. Or we could use this as weights during the fitting of a variogram function.

[...]

PR #131, opened 2021-01-14
Automatic Binning

MuellerSeb:
Closes: #55 #36 

This PR adds a simple automatic binning routine `standard_bins` to the variogram submodule.
It returns an equal division from 0 to `max_dist` with `bin_no` as number of bins. The following rules are applied:
- if `max_dist` is not given, it is set to one third of the box diameter
- if `bin_no` is not given, it is determined by sturges' rule from the number of point combinations

You can now call `gs.vario_estimate` without passing `bin_edges`. In this case, the bin_edges will be determined as described above.


MuellerSeb:
@LSchueler ready for review. :blush:

examples/03_variogram/06_auto_bin_latlon.py
```diff
@@ -0,0 +1,76 @@
...
+ax = krige.plot()
+# plotting lat on y-axis and lon on x-axis
+ax.scatter(pos[1], pos[0], 50, c=field, edgecolors="k", label="input")
+ax.legend()
+
+###############################################################################
+# Looks good, doesn't it?
+#
+# This example shows, that setting up variogram estimation and kriging routines
+# is straight forward with GSTools. ;-)
```
LSchueler, line 76: Really nice example!

gstools/variogram/binning.py
```diff
@@ -0,0 +1,100 @@
...
+
+__all__ = ["standard_bins"]
+
+
+def _sturges(pnt_cnt):
+    return int(np.ceil(2 * np.log2(pnt_cnt) + 1))
+
+
+def _chordal_to_great_circle(dist):
+    return 2 * np.arcsin(np.minimum(np.maximum(np.divide(dist, 2), 0), 1))
```
LSchueler, line 30: I think this could be put into `gstools/tools`.

LSchueler (changes requested):
This is a really nice convenience addition!

What do you think of moving the `_chordal_to_great_circle` method to a more visible place and making it public?

Appart from that, I think we are ready to merge this PR.

LSchueler:
I fixes some typos, so don't forget to pull.

examples/03_variogram/06_auto_bin_latlon.py
MuellerSeb: :blush:

MuellerSeb:
Strange. It was merged, but this PR says it has conflicts and can't be merged. It felt like an internal GitHub Error when I clicked on "merge" and I had to reload the page. Hope nothing was broken.

PR #124, opened 2020-12-21
Mean, Normalizer, Trend and Fitting

MuellerSeb:
Closes: #94 
# Abstract
This PR adds:
- a new submodule containing normalizer classes.
- a callable mean (added to normalized data) for all Field classes
- a callable trend (added to again-transformed / input data) for all Field classes
- normalizer as input for all field classes
- auto fit option in Krige class (normalizer and variogram)

##  Purpose of Normalizer
Their purpose is to normalize (follow a normal distribution) and denormalize data.
Since Krige and SRF produce normal distributed fields, these normalizers offer the opportunity to transform fields into a desired distribution and, in case of krige, to normalize conditioning data beforehand.

## Normalizer
- LogNormal
- BoxCox
- BoxCoxShift
- YeoJohnson
- Modulus
- Manly

## Normalizer Fitting
In order to fit hyper-parameter of the normalization transformations, the Normalize base-class provides a `fit` method, that estimates parameters by maximizing the log-likelihood function for given data.

## Callable trend and mean
In order to use trends and variable means from external source to play nicely together with Normalizer, we will allow callable means to be added to normalized data and trends applied to input data.

## Auto Fitting for Krige
SInce we provide an automated workflow for estimating the empirical variogram with automatic binning, we plug that into the kriging class as an automated workflow. In addition the normalizer can be fitted automatically as well by minimizing the likelihood function as described above. 

## Affected Routines
- new method `Field.post_field` (replacing `Krige._post_field` and is similar to `Field.pre_pos`)
- `Krige._krige_cond` needs to care about normalizer and callable mean in correct order
- `Field.__init__` needs `trend_function`, `normalizer` and `fit_normalizer` arguments
- add `Krige.cond_mean` holding mean values at conditioning points

# TODO
- [x] add to all Fields as keywords
- [x] add examples
- [x] add tests

PR #134, opened 2021-01-16
Replace pyvista 3d plot with matplotlib animation in Precip example

LSchueler:
I've updated the precipitation example to now show the time evolution of the 2D P-field as an animated figure.

### DIFF
--- a/examples/00_misc/03_precipitation.py
+++ b/examples/00_misc/03_precipitation.py
@@ -15,6 +15,7 @@ series.
 import copy
 import numpy as np
 import matplotlib.pyplot as plt
+import matplotlib.animation as ani
 import gstools as gs
 
 # fix the seed for reproducibility
@@ -149,9 +150,17 @@ amount = 3.0
 srf.field *= amount
 
 ###############################################################################
-# plot the 2d precipitation field together with the time axis as a 3d plot.
-# We will cut out the volumes with low precipitation in order to have a look
-# inside the cuboid.
+# plot the 2d precipitation field over time as an animation.
 
-mesh = srf.to_pyvista()
-mesh.threshold_percent(0.25).plot()
+def _update_ani(idx):
+    quad.set_array(srf.field[idx, :, :].T.ravel())
+    return quad,
+
+fig, ax = plt.subplots()
+quad = ax.pcolormesh(x, y, srf.field[0,:,:].T, cmap="Blues", shading="gouraud")
+cbar = fig.colorbar(quad)
+cbar.ax.set_ylabel(r"Precipitation $P$ / mm")
+ax.set_xlabel(r"$x$ / km")
+ax.set_ylabel(r"$y$ / km")
+
+ani = ani.FuncAnimation(fig, _update_ani, len(t), interval=100, blit=True)
--- a/examples/09_spatio_temporal/02_precip_2d.py
+++ b/examples/09_spatio_temporal/02_precip_2d.py
@@ -1,0 +1,74 @@
+"""
+Creating a 2D Synthetic Precipitation Field
+-------------------------------------------
+
+In this example we'll create a time series of a 2D synthetic precipitation
+field.
+
+Very similar to the previous tutorial, we'll start off by creating a Gaussian
+random field with an exponential variogram, which seems to reproduce the
+spatial correlations of precipitation fields quite well. We'll create a daily
+timeseries over a two dimensional domain of 50km x 40km. This workflow is
+suited for sub daily precipitation time series.
+"""
+
+import numpy as np
+import matplotlib.pyplot as plt
+import matplotlib.animation as animation
+import gstools as gs
+
+# fix the seed for reproducibility
+seed = 20170521
+# 1st spatial axis of 50km with a resolution of 1km
+x = np.arange(0, 50, 1.0)
+# 2nd spatial axis of 40km with a resolution of 1km
+y = np.arange(0, 40, 1.0)
+# half daily timesteps over three months
+t = np.arange(0.0, 90.0, 0.5)
+
+# total spatio-temporal dimension
+st_dim = 2 + 1
+# space-time anisotropy ratio given in units d / km
+st_anis = 0.4
+
+# an exponential variogram with a corr. lengths of 5km, 5km, and 2d
+model = gs.Exponential(dim=st_dim, var=1.0, len_scale=5.0, anis=st_anis)
+# create a spatial random field instance
+srf = gs.SRF(model, seed=seed)
+
+pos, time = [x, y], [t]
+
+# the Gaussian random field
+srf.structured(pos + time)
+
+# the dry periods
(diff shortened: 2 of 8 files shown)


MuellerSeb (changes requested):
This animation look amazing! Thanks for adding!

I had a second though on spatio-temporal modeling:
We should create a separate example folder/category for "Spatio-Temporal Modelling".

Precipitation is one of the prime application examples, so your work could shine here :wink:
But I'd split your example in two, one 1D and the other one 2D.

In order to make the spatio-temporal people happy, I'd suggest to come up with some conventions:

1. What we provide at the moment is a so called *"metric"* spatio-temporal model. We should mention that.
2. time dimension should be the last one:
    - common in the spatio-temporal literature
    - Since `anis` in the `CovModel` will be filled up from the right, a passed `st_anis` for the space-time anisotropy of a metric model will be used correctly: (lucky we made a good decision here)
        ```python
        st_dim = 2 + 1  # spatial dimension of two and one time dimension
        st_anis = 0.4  # space-time anisotropy given in d / km
        # an exponential variogram with a corr. lengths of 5km, 5km and 2d
        model = gs.Exponential(dim=st_dim, var=1.0, len_scale=5.0, anis=st_anis)
        ```
    - since we now provide arbitrary dimensions with the special definition of rotation angles, the first rotation angles for the first n dimension in a (n+1) dimension model coincide with the rotation angles of a n-dimensional model (lucky we made a good decision here as well) .. this means 3D spatio-temporal models work out of the box :wink:
3. We should again use the `pos` tuple/list like in the other examples. We can use another list for the time extension to make it more visible what happens:
    ```python
    pos, time = [x, y], [t]
    srf = gs.SRF(model, seed=seed)
    srf.structured(pos + time)
    ```
    PS: The `seed` should be passed to `SRF` directly, so we the modes are not sampled twice in this example.
    We can use this `pos` later on `ax.pcolormesh(*pos, srf.field ...`
    
What do you think about that?

examples/00_misc/03_precipitation.py
```diff
@@ -149,9 +150,17 @@
...
+# plot the 2d precipitation field over time as an animation.
 
-mesh = srf.to_pyvista()
-mesh.threshold_percent(0.25).plot()
+def _update_ani(idx):
+    quad.set_array(srf.field[idx, :, :].T.ravel())
+    return quad,
+
+fig, ax = plt.subplots()
+quad = ax.pcolormesh(x, y, srf.field[0,:,:].T, cmap="Blues", shading="gouraud")
```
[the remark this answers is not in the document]
LSchueler: I never really liked `imshow`. I don't know why, but consistency is a good argument.

examples/00_misc/03_precipitation.py
```diff
@@ -149,9 +150,17 @@
...
+cbar = fig.colorbar(quad)
+cbar.ax.set_ylabel(r"Precipitation $P$ / mm")
+ax.set_xlabel(r"$x$ / km")
+ax.set_ylabel(r"$y$ / km")
```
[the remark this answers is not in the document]
LSchueler: YEAH!!!

examples/00_misc/03_precipitation.py
```diff
@@ -149,9 +150,17 @@
...
-# plot the 2d precipitation field together with the time axis as a 3d plot.
-# We will cut out the volumes with low precipitation in order to have a look
-# inside the cuboid.
+# plot the 2d precipitation field over time as an animation.
 
-mesh = srf.to_pyvista()
-mesh.threshold_percent(0.25).plot()
+def _update_ani(idx):
+    quad.set_array(srf.field[idx, :, :].T.ravel())
+    return quad,
```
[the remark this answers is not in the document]
LSchueler: Well, to give users a hint that it's not an `imshow` object ;-)

LSchueler:
Thanks for your input! Having had more to do with special relativity than with spatio-temporal geostatistics I'm not used to the notion of putting the time dimension at the end of the space-time-vector.

I've split and updated the example accordingly, but I didn't have time to include any introductory text yet.

On my local machine, the matplotlib animation is only shown as a static image in the sphinx-gallery. Any idea?

examples/09_spatio_temporal/02_precip_2d.py
```diff
@@ -0,0 +1,74 @@
...
+
+# the Gaussian random field
+srf.structured(pos + time)
+
+# the dry periods
+threshold = 0.4
+srf.field[srf.field <= threshold] = 0.0
+
+# account for the skewness
+gs.transform.boxcox(srf, lmbda=0.5, shift=-1.0)
```
MuellerSeb, line 49: I think you can drop the threshold line. These `0.0` aren't zero after the boxcox transformation and this creates strange ouput in terms of precipitation (same problem in the 1D example).

The boxcox transformation is implemented to cut off values, that can't be transformed, so you should result in "realistic" values when solely applying the boxcox trafo.

MuellerSeb (changes requested):
Like it much more! :wink:
Only thing is the threshold setting, which doesn't work well with the boxcox transformation afterwards.

And a `README.rst` is missing in the new examples folder. The following should be sufficient:
```rst
Spatio-Temporal Modelling
=========================

Examples
--------
```

With this, the documentation compiles correctly and the animation is working.

examples/09_spatio_temporal/02_precip_2d.py
MuellerSeb: Or you do the following:
```python
lmbda = 0.5
gs.transform.boxcox(srf, lmbda=lmbda, shift=-1 / lmbda)
```

MuellerSeb: Or we also implement the yeo-johnson transformation in the `transform` submodule

LSchueler:
> Like it much more! wink
> Only thing is the threshold setting, which doesn't work well with the boxcox transformation afterwards.
> 
> And a `README.rst` is missing in the new examples folder. The following should be sufficient:
> 
> ```
> Spatio-Temporal Modelling
> =========================
> 
> Examples
> --------
> ```
> 
> With this, the documentation compiles correctly and the animation is working.

Yeah... I forgot to add the readme to the commit stage. But that means that something is not working with sphinx gallery on my local machine. Whatever, as long as it runs on RTD.

examples/09_spatio_temporal/02_precip_2d.py
LSchueler: Good catch, thanks!

LSchueler:
I've updated the examples accordingly.

examples/09_spatio_temporal/02_precip_2d.py
MuellerSeb: When #124 is merged, we can also rewrite that to use the box-cox normalizer.

MuellerSeb (approved):
Yeah!

PR #136, opened 2021-01-27
Examples: update dwd example to use wetterdienst 0.13

MuellerSeb:
Simple update to reduce runtime.


LSchueler (approved):
Have to get rid of those pesky hard-coded values, nice ;-)

PR #137, opened 2021-02-12
Add example of how to plot a 3d vector field

LSchueler:
In this short example I demonstrate how to use PyVista to create an external mesh on which a 3d vector field is generated.
I'm not sure why the colours look a but dull in Sphinx-Gallery. When looking at the plot in PyVista, the colours are more vibrant. But this is not a big issue I guess.


examples/04_vector_field/01_3d_vector_field.py
```diff
@@ -0,0 +1,70 @@
...
+
+In this example we are going to generate a random 3D vector field with a
+Gaussian covariance model. The mesh on which we generate the field will be
+externally defined and it will be generated by PyVista.
+"""
+import numpy as np
+import gstools as gs
+import pyvista as pv
+
+pv.set_plot_theme("document")
```
MuellerSeb, line 13: Why this?

examples/04_vector_field/01_3d_vector_field.py
```diff
@@ -0,0 +1,70 @@
...
+
+###############################################################################
+# create a uniform grid with PyVista
+nx, ny, nz = 40, 30, 10
+mesh = pv.UniformGrid((nx, ny, nz), (1., 1., 1.), (-10., 0., 0.))
```
MuellerSeb, line 18: would prefere:
```python
dim, spacing, origin = (40, 30, 10), (1, 1, 1), (-10, 0, 0)
mesh = pv.UniformGrid(dim, spacing, origin)
```
Following the documentation of pyvista.

examples/04_vector_field/01_3d_vector_field.py
```diff
@@ -0,0 +1,70 @@
...
+model = gs.Gaussian(dim=3, var=3, len_scale=1.5)
+srf = gs.SRF(model, generator='VectorField')
+srf((x, y, z), mesh_type='unstructured', seed=198412031)
+
+# add a mean velocity in x-direction
+srf.field[0,:] += 0.5
+
+###############################################################################
+# add the velocity field to the mesh object
+mesh["Velocity"] = srf.field.T
```
MuellerSeb, line 34: Why not using the `mesh` method? And set seed in `__init__` so sampling is done only once,
```python
srf = gs.SRF(model, generator='VectorField', seed=198412031)
srf.mesh(mesh, points="points", name="Velocity")
mesh["Velocity"][:, 0] += 0.5
```

examples/04_vector_field/01_3d_vector_field.py
```diff
@@ -0,0 +1,70 @@
+"""
+Generating a Random 3D Vector Field
+-----------------------------------
+
+In this example we are going to generate a random 3D vector field with a
+Gaussian covariance model. The mesh on which we generate the field will be
+externally defined and it will be generated by PyVista.
+"""
+import numpy as np
```
MuellerSeb, line 9: missing thumbnail in gallery:
```python
# sphinx_gallery_thumbnail_path = 'pics/GS_pyvista_streamlines.png'
```

MuellerSeb (changes requested):
That's awesome! Some minor things from my side in the review.
And: I would rename `mesh` to `grid` following the pyvista documentation on `UniformGrid`.
Also the thumbnail is missing. Have a look at the other pyvista example for this.

examples/04_vector_field/01_3d_vector_field.py
LSchueler: clarified in example

examples/04_vector_field/01_3d_vector_field.py
LSchueler: Good catch, thanks!

examples/04_vector_field/01_3d_vector_field.py
LSchueler: Yeah, that makes things clearer.

LSchueler:
> That's awesome! Some minor things from my side in the review.
> And: I would rename `mesh` to `grid` following the pyvista documentation on `UniformGrid`.
> Also the thumbnail is missing. Have a look at the other pyvista example for this.

Yeah, that seems to be a bit inconsistent in PyVista, I followed the example [here](https://docs.pyvista.org/examples/01-filter/streamlines.html#custom-3d-vector-field) and I call the method `Plotter.add_mesh()` in this example, so I think it's the least confusing if I stick to `mesh`. And `SRF` also has a `mesh()` method ;-) .

examples/04_vector_field/01_3d_vector_field.py
LSchueler: :-D I'm simply not keeping up with the development of GSTools... But I think the method is not working with vector fields... I'm gonna test that again and open an issue if that is true.
I suggest to keep the example as it is for the moment and update it, as soon as I figured out the problem.

MuellerSeb: You are right... it's not working with Vector fields. Damn.. please create an issue.

examples/04_vector_field/01_3d_vector_field.py
MuellerSeb: Cool!

examples/04_vector_field/01_3d_vector_field.py
```diff
@@ -0,0 +1,73 @@
...
+z = mesh.points[:, 2]
+
+###############################################################################
+# create an incompressible random 3d velocity field on the given mesh
+model = gs.Gaussian(dim=3, var=3, len_scale=1.5)
+srf = gs.SRF(model, generator='VectorField', seed=198412031)
+srf((x, y, z), mesh_type='unstructured')
+
+# add a mean velocity in x-direction
+srf.field[0, :] += 0.5
```
MuellerSeb, line 33: Couldn't this be provided by a mean? I think the new mean-normalizer-trend thingy is capable of dealing with vectors. But this is just fine here ;-)

MuellerSeb (approved):
I fine with this fine example :relaxed:

examples/04_vector_field/01_3d_vector_field.py
LSchueler: Will do, but I won't make it today anymore...

examples/04_vector_field/01_3d_vector_field.py
LSchueler: I'll update the example when the normalizer is ready. This example is gonna show a lot of fancy things ;-)

MuellerSeb:
Info:
vector valued mean added: https://github.com/GeoStat-Framework/GSTools/pull/124/commits/9bb8daafc08880eac38a15c3b86a0e540bcf9785
mesh call for vector valued fields fixed: https://github.com/GeoStat-Framework/GSTools/pull/124/commits/56ec497e2d52079c55f74ff8012d0fa9876e7661

PR #146, opened 2021-03-22
Drop Python 3.5 support and update CI

MuellerSeb:
Python 3.5 had its EOL in September 2020 and Ubuntu 16.04 LTS (where Python 3.5 was installed by default) has its End of Standard Support in April 2021. Let's drop support now.

Changes:
- Use `ModuleNotFoundError` when version file could not be found (was not present in py35)
- remove metaclass hack for py35 in `CovModel`
- Use new Github action for `cibuildwheel`
- update setup requirement versions for wheels built: `numpy==1.19.* cython==0.29.*`
- pin version for pypi gh-action

### DIFF
--- a/gstools/covmodel/base.py
+++ b/gstools/covmodel/base.py
@@ -24,7 +24,6 @@ from gstools.tools.geometric import (
     pos2latlon,
 )
 from gstools.covmodel.tools import (
-    InitSubclassMeta,
     _init_subclass,
     set_opt_args,
     set_len_anis,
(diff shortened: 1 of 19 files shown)


MuellerSeb:
Windows wheels still failing because https://github.com/GeoStat-Framework/GSTools/pull/147 is not yet merged.

gstools/covmodel/base.py
```diff
@@ -47,7 +46,7 @@
 HANKEL_DEFAULT = {"a": -1, "b": 1, "N": 200, "h": 0.001, "alt": True}
 
 
-class CovModel(metaclass=InitSubclassMeta):
+class CovModel(metaclass=type):
```
MuellerSeb, line 49: @LSchueler is this the way to do it correctly?

MuellerSeb: OK, metaclass was now redundant. We needed this, since `__init_subclass__` was added in py36: https://docs.python.org/3/reference/datamodel.html#object.__init_subclass__

LSchueler (approved):
Bye bye annoying Py3.5 and welcome f-strings!!

PR #147, opened 2021-03-22
CovModels: limit exponent shape parameters to 50 to prevent numerical errors

MuellerSeb:
Some models with shape parameters could run into numerical errors during variogram fitting, since their shape parameters had no upper bound.

Namely:
- TPLSimple
- SuperSpherical
- JBessel
- Rational

The new default upper bound for the shape parameter of these models is `50`. (Can still be changed by the user)


MuellerSeb:
CI currently failing for linux wheels because of this: https://status.quay.io/

LSchueler (approved):
Let's get this thing stable!

PR #154, opened 2021-04-06
Add pyproject.toml

MuellerSeb:
This PR adds a pyproject.toml and updates the installation setup.
- rely on pep517/518 builds for wheels and sdist
- removed all `requirements*.txt` files (in setup.cfg now)
- use `setup.cfg` for all setup configurations
- use `setup.py` only for cython and openmp configurations
- remove install options for openmp and use an environment variable `GSTOOLS_BUILD_PARALLEL` (like sklearn) instead
- use `pyproject.toml` to define build dependencies, setuptools_scm, coverage and black
- add format checking stage in CI (using black)
- update CI script to always use pip and the extras defined in setup.cfg (`[test]`)
- add `.readthedocs.yml` to install gstools with pip and `[doc]` extras on readthedocs
- updated `MANIFEST.in` to shrink wheels and sdist (from 45mb to 28mb in the artifacts on GH-Actions)
- removed development extra `[dev]` from setup which is equal to `pip install gstools[test, doc]`


LSchueler (approved):
I only had a quick look, but this looks so much cleaner! Approved! ;-)

PR #157, opened 2021-04-08
Covmodel arg bugfix: Always convert nugget to float

MuellerSeb:
This line of code generated an Error with `gstools==1.3.0rc2`:
```python
import gstools as gs
print(gs.Matern(nugget=0))
```
```bash
ValueError: Precision not allowed in integer format specifier
```

Reason: `nugget` was never converted to `float` internally and is stored as `int` (which is totally fine for all calculations), but the printing routine tries to format it as float.

This PR solves this problem and adds some minor additions:
- `rescale` was not setable after the model was instantiated: added `rescale.setter`
- `rescale` didn't show up in the `repr` string: fixed if rescale is not the default value
- `anis` and `angles` where always showing up also for isotropic and unrotated models: fixed

This means, the repr is the exact init sequence to create a model:
```python
from gstools import Matern
exec(f"m={Matern()}")
print(m)
```
```bash
Matern(dim=3, var=1.0, len_scale=1.0, nugget=0.0, nu=1.0)
```
The conditional repr looks like the following:
```python
from gstools import Matern
print(Matern(rescale=2))
print(Matern(anis=2))
print(Matern(angles=1))
print(Matern(rescale=2, anis=2, angles=1))
```
```bash
Matern(dim=3, var=1.0, len_scale=1.0, nugget=0.0, rescale=2.0, nu=1.0)
Matern(dim=3, var=1.0, len_scale=1.0, nugget=0.0, anis=[1.0, 2.0], nu=1.0)
Matern(dim=3, var=1.0, len_scale=1.0, nugget=0.0, angles=[1.0, 0.0, 0.0], nu=1.0)
Matern(dim=3, var=1.0, len_scale=1.0, nugget=0.0, anis=[1.0, 2.0], angles=[1.0, 0.0, 0.0], rescale=2.0, nu=1.0)
```

### DIFF
--- a/gstools/covmodel/tools.py
+++ b/gstools/covmodel/tools.py
@@ -576,11 +576,19 @@ def model_repr(model):  # pragma: no cover
     model : :any:`CovModel`
         The covariance model in use.
     """
-    opt_str = ""
+    if not np.isclose(model.rescale, model.default_rescale()):
+        opt_str = ", rescale={0:.{p}}".format(model.rescale, p=model._prec)
+    else:
+        opt_str = ""
     for opt in model.opt_arg:
-        opt_str += (
-            ", " + opt + "={0:.{1}}".format(getattr(model, opt), model._prec)
+        opt_str += ", {0}={1:.{p}}".format(
+            opt, getattr(model, opt), p=model._prec
         )
+    # only print anis and angles if model is anisotropic or rotated
+    ani = "" if model.is_isotropic else f", anis={list_format(model.anis, 3)}"
+    ang = (
+        f", angles={list_format(model.angles, 3)}" if model.do_rotation else ""
+    )
     if model.latlon:
         std_str = (
             "{0}(latlon={1}, var={2:.{p}}, len_scale={3:.{p}}, "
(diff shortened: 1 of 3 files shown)


gstools/covmodel/tools.py
```diff
@@ -576,11 +576,19 @@ def model_repr(model):  # pragma: no cover
...
-    opt_str = ""
+    if not np.isclose(model.rescale, model.default_rescale()):
+        opt_str = ", rescale={0:.{p}}".format(model.rescale, p=model._prec)
+    else:
+        opt_str = ""
     for opt in model.opt_arg:
-        opt_str += (
-            ", " + opt + "={0:.{1}}".format(getattr(model, opt), model._prec)
+        opt_str += ", {0}={1:.{p}}".format(
+            opt, getattr(model, opt), p=model._prec
```
LSchueler, line 585: I think
`opt_str += f", {opt}={getattr(model, opt):.{model._prec}}"`
looks much cleaner, f-strings WTF!

gstools/covmodel/tools.py
```diff
@@ -597,14 +605,14 @@ def model_repr(model):  # pragma: no cover
...
+            "nugget={4:.{p}}{5}{6}{7})".format(
                 model.name,
                 model.dim,
                 model.var,
                 model.len_scale,
                 model.nugget,
-                list_format(model.anis, 3),
-                list_format(model.angles, 3),
+                ani,
+                ang,
```
LSchueler, line 615: Hmm, not so sure about f-strings here... :laughing:

LSchueler (approved):
Funny little bug :-)
Thanks for taking a look it.
Check the f-strings and then you can happily merge!

gstools/covmodel/tools.py
MuellerSeb: You're right! f-strings FTW!

gstools/covmodel/tools.py
MuellerSeb: Done. Was worth the effort.

Issue #24, opened 2019-07-17
At risk of circular imports


MuellerSeb:
Since all imports within the package are absolute we are at risk to run into circular imports when something is changed.

See for example this bug-fix: https://github.com/GeoStat-Framework/GSTools/commit/73ba330642261f9a59a955dbd82a1037d78edb89

Possibilities:
https://stackoverflow.com/questions/7336802/how-to-avoid-circular-imports-in-python

LSchueler:
Did I get you correctly, that you are afraid to run into a problem sometime in future and the problem only affects Python 2, which will be deprecated in the near future?

Source from one of the answers of the link you provided:

"The ImportError and AttributeError issues only occur in python 2. In python 3 the import machinery has been rewritten and all of these import statements (with the exception of 4) will work, even with circular dependencies."

I suggest sitting back, having a cup of coffee and waiting for Python 2 to finally vanish ;-)

MuellerSeb:
I ran into that problem and I am using Python3 .. I was actually a bit confused by this statement in the stackoverflow conversation.
Just try loading gstools from commit https://github.com/GeoStat-Framework/GSTools/commit/f2f6c084fcec018aee663bf16f7dc8f2e1c86c8e

MuellerSeb:
Since we now check the source code with pylint we should be safe here!

PR #159, opened 2021-04-08
Pylint fixes

MuellerSeb:
... I've done it. Pylint is a pain, but a good one.

I created a setup for Pylint in `pyproject.toml` that should warn us about bad practice. I added a new stage in CI to check pylint.

Certain Pylint rules are disabled per file, where it makes sence (at least for me).

It helped me figure out all of the circular imports (see #24). Especially the matplotlib imports are now in a try block and later on deferred.

All remaining pylint warnings were sorted out here.

### DIFF
--- a/gstools/field/base.py
+++ b/gstools/field/base.py
@@ -9,7 +9,7 @@ The following classes are provided
 .. autosummary::
    Field
 """
-# pylint: disable=C0103, W0107, C0415
+# pylint: disable=C0103, C0415
 from functools import partial
 import numpy as np
 from gstools.covmodel.base import CovModel
--- a/gstools/field/base.py
+++ b/gstools/field/base.py
@@ -9,7 +9,7 @@ The following classes are provided
 .. autosummary::
    Field
 """
-# pylint: disable=C0103, W0107
+# pylint: disable=C0103, W0107, C0415
 from functools import partial
 import numpy as np
 from gstools.covmodel.base import CovModel
--- a/gstools/field/base.py
+++ b/gstools/field/base.py
@@ -9,7 +9,7 @@ The following classes are provided
 .. autosummary::
    Field
 """
-# pylint: disable=C0103
+# pylint: disable=C0103, W0107
 from functools import partial
 import numpy as np
 from gstools.covmodel.base import CovModel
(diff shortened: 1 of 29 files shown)


MuellerSeb:

gstools/field/base.py
```diff
@@ -297,9 +296,8 @@ def plot(
                 "Specify 'scalar' or 'vector' before plotting."
             )
 
-        elif self.value_type == "scalar":
+        if self.value_type == "scalar":
```
LSchueler, line 299: Is some super deep Python knowledge required to understand why elif is wrong here? Or is this a convention to separate error raising from other branching logic?

LSchueler (commented):
I just had a very very brief look at this PR. Some changes I could not comprehend so quickly, but most of them look pretty harmless.
Maybe it would be helpful to have one central file (some source code file) where the pylint disabling codes are summarized or has the documentaion gotten better over time? - The last time I was setting the linter, it was a real pain to find the correct codes.
Actually, I think I would prefer a description on why each setting was made for each file. I'm just not sure how cluttered the files would be.

But anyway, I'm really curious to see what the Linter will complain about in my future code :-)

gstools/field/base.py
MuellerSeb: See here: https://vald-phoenix.github.io/pylint-errors/plerr/errors/refactoring/R1720

MuellerSeb: The rule simply says: no need for an `else` after a `raise`.

MuellerSeb:
@LSchueler there is this Project: https://vald-phoenix.github.io/pylint-errors/ where the error-codes are documented quite well. I added a section in `CONTRIBUTING.md` about the occasionally disabled pylint checks. Hope you like it.

LSchueler:
> @LSchueler there is this Project: https://vald-phoenix.github.io/pylint-errors/ where the error-codes are documented quite well. I added a section in `CONTRIBUTING.md` about the occasionally disabled pylint checks. Hope you like it.

That's great, I like the summary and also that you can find it in `CONTRIBUTING.md`.

LSchueler (approved):

Issue #161, opened 2021-04-12
No access to fitted normalizer from variogram estimation


MuellerSeb:
If using `fit_normalizer=True` during variogram estimation, one can't access the fitted normalizer afterwards (in v1.3.0rc2):
```python
bin_center, gamma = gs.vario_estimate((x, y), field, normalizer=gs.normalizer.BoxCox, fit_normalizer=True)
```
It should return the fitted normalizer as well, when fitting it:
```python
bin_center, gamma, normalizer = gs.vario_estimate((x, y), field, normalizer=gs.normalizer.BoxCox, fit_normalizer=True)
```
In order to use it afterwards in `SRF` or `Krige`.

PR #162, opened 2021-04-12
return fitted normalizer

MuellerSeb:
This PR adds optional return of `normalizer` in `vario_estimate` when `fit_normalizer=True`.

Closes: #161

### DIFF
--- a/gstools/normalizer/tools.py
+++ b/gstools/normalizer/tools.py
@@ -77,7 +77,7 @@ def apply_mean_norm_trend(
 
     Returns
     -------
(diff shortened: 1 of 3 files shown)


MuellerSeb:

LSchueler (approved):
This is a nice little addition towards better practicability of the variogram estimation, thanks!

I'm wondering if things are getting so complex with all the different return branches, that this should somehow be covered by the unit tests or if this is still manageable in the good old fassioned way, by simply leaning back and taking a sharp look at it... ;-)

PR #165, opened 2021-04-17
build: use oldest-supported-numpy

### LINKED ISSUE #164: DLL load failed
Hi,

I installed gstools via
pip install gstools
it installed from gstools-1.3.0-cp36-cp36m-win_amd64.whl
but when I do
import gstools
it ends with
from gstools.variogram.estimator import (
ImportError: DLL load failed: The specified module could not be found.
went back to gstools 1.2.1 and all workds fine

Thank you
Peter

MuellerSeb:
Closes #164 

Use the meta package for numpy as a build dep: https://github.com/scipy/oldest-supported-numpy

See:
https://scikit-hep.org/developer/packaging#special-additions-numpy

The reason to use the oldest available Numpy version as a build-time dependency is because of ABI compatibility. Binaries compiled with old Numpy versions are binary compatible with newer Numpy versions, but not vice versa.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -3,7 +3,7 @@ requires = [
     "setuptools>=42",
     "wheel",
     "setuptools_scm[toml]>=3.5",
-    "numpy>=1.14.5,<2.0",
+    "oldest-supported-numpy",
     "Cython>=0.28.3,<3.0",
 ]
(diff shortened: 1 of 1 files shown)


LSchueler (approved):
That's a nice and easy fix, thanks for handling this!

Issue #61, opened 2020-01-12
Handling pre-built structured grids


banesullivan:
the structured methods in this library (to my knowledge) only support 1-D arrays representing the axial coordinates of a rectilinear grid (the inputs to the `numpy.meshgrid`) for structured points. This is okay... but in cases of curvilinear structured grids where the user has their own tooling for building those grids, we may want to pass prebuilt coordinate arrays.

For example, take this curvilinear structured grid:  [curvi_mesh.vts.zip](https://github.com/GeoStat-Framework/GSTools/files/4051278/curvi_mesh.vts.zip)

```py
import pyvista as pv

mesh = pv.read("curvi_mesh.vts")

mesh.plot(show_edges=True, show_grid=True)
```
![download](https://user-images.githubusercontent.com/22067021/72224577-3366bb00-3539-11ea-8f53-03465ab1ad62.png)

This type of grid cannot be made with NumPy's `meshgrid` function as it cannot be represented as 3 1D coordinate arrays. This is a set of structured coordinates just like what `numpy.meshgrid` yields. These structured coordinates can be accessed by the `x`, `y`, and `z` attributes of a PyVista `StructuredGrid` object. 

Note that this is something that is really common outside of PyVista too - take [`discretize`'s `CurvilinearMesh` class for instance](http://discretize.simpeg.xyz/en/master/api/generated/discretize.CurvilinearMesh.html#discretize.CurvilinearMesh).

```py
>>> mesh.x.shape, mesh.y.shape, mesh.z.shape
((17, 17, 5), (17, 17, 5), (17, 17, 5))
```

So how can we pass these kinds of pre-built structured points to the field classes such that we can leverage any performance/optimization for structured grids and plotting routines?

Just to drive home that this is a pretty common thing, even matplotlib can handle these types of structured coordinates:

```py
plt.pcolormesh(mesh.x[:,:,0], 
               mesh.y[:,:,0], 
               np.random.rand(*mesh.dimensions[0:2]),
               color="k")
plt.axis("image")
plt.show()
```

![download](https://user-images.githubusercontent.com/22067021/72224752-13d09200-353b-11ea-9b70-8f0a34124dbd.png)

-----

So basically, how might I use that structured gird with GSTools without just sending the points as an unstructured pointset. e.g.

```py
from gstools import SRF, Gaussian

model = Gaussian(dim=2, var=0.2, len_scale=0.05)
srf = SRF(model, seed=20170519)

[...]

PR #166, opened 2021-04-18
Field: enable standalone use

MuellerSeb:
Minor update to the `Field` class to enable standalone usage: 
- enable standalone use with given `dim`
- add `points` method to generate field on a point list (transposed unstructured pos tuple as used in many other packages (e.g. PyVista))
- add `latlon` property
- better doc ref to `__call__`

This was a low hanging fruit. Now `Field` can be used to display and export arbitrary n-d point data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 All notable changes to **GSTools** will be documented in this file.
 
+## [1.3.1] - Pure Pink - 2021-?
+
+### Enhancements
+- Standalone use of Field class [#166](https://github.com/GeoStat-Framework/GSTools/issues/166)
+- add social badges in README [#169](https://github.com/GeoStat-Framework/GSTools/issues/169), [#170](https://github.com/GeoStat-Framework/GSTools/issues/170)
+
+### Bugfixes
+- use `oldest-supported-numpy` to build cython extensions [#165](https://github.com/GeoStat-Framework/GSTools/pull/165)
+
+
 ## [1.3.0] - Pure Pink - 2021-04
 
 ### Topics
--- a/examples/00_misc/05_standalone_field.py
+++ b/examples/00_misc/05_standalone_field.py
@@ -1,0 +1,27 @@
+"""
+Standalone Field class
+----------------------
+
+The :any:`Field` class of GSTools can be used to plot arbitrary data in nD.
+
+In the following example we will produce 10000 random points in 4D with
+random values and plot them.
+"""
+import numpy as np
+import gstools as gs
+
+x1 = np.random.RandomState(19970221).rand(10000) * 100.0
+x2 = np.random.RandomState(20011012).rand(10000) * 100.0
+x3 = np.random.RandomState(20210530).rand(10000) * 100.0
+x4 = np.random.RandomState(20210531).rand(10000) * 100.0
+values = np.random.RandomState(2021).rand(10000) * 100.0
+
+###############################################################################
+# Only thing needed to instantiate the Field is the dimension.
+#
+# Afterwards we can call the instance like all other Fields
+# (:any:`SRF`, :any:`Krige` or :any:`CondSRF`), but with an additional field.
+
+plotter = gs.field.Field(dim=4)
+plotter(pos=(x1, x2, x3, x4), field=values)
+plotter.plot()
--- a/examples/00_misc/05_standalone_field.py
+++ b/examples/00_misc/05_standalone_field.py
@@ -10,11 +10,12 @@ random values and plot them.
 import numpy as np
 import gstools as gs
 
-x1 = np.random.RandomState(19970221).rand(10000) * 100.0
-x2 = np.random.RandomState(20011012).rand(10000) * 100.0
-x3 = np.random.RandomState(20210530).rand(10000) * 100.0
-x4 = np.random.RandomState(20210531).rand(10000) * 100.0
-values = np.random.RandomState(2021).rand(10000) * 100.0
+rng = np.random.RandomState(19970221)
+x0 = rng.rand(10000) * 100.0
+x1 = rng.rand(10000) * 100.0
+x2 = rng.rand(10000) * 100.0
+x3 = rng.rand(10000) * 100.0
+values = rng.rand(10000) * 100.0
 
 ###############################################################################
 # Only thing needed to instantiate the Field is the dimension.
@@ -23,5 +24,5 @@ values = np.random.RandomState(2021).rand(10000) * 100.0
 # (:any:`SRF`, :any:`Krige` or :any:`CondSRF`), but with an additional field.
 
 plotter = gs.field.Field(dim=4)
-plotter(pos=(x1, x2, x3, x4), field=values)
+plotter(pos=(x0, x1, x2, x3), field=values)
 plotter.plot()
--- a/gstools/field/base.py
+++ b/gstools/field/base.py
@@ -136,12 +136,12 @@ class Field:
     def mesh(
         self, mesh, points="centroids", direction="all", name="field", **kwargs
     ):
-        """Generate a field on a given meshio or ogs5py mesh.
+        """Generate a field on a given meshio, ogs5py or PyVista mesh.
 
         Parameters
         ----------
         mesh : meshio.Mesh or ogs5py.MSH or PyVista mesh
-            The given meshio, ogs5py, or PyVista mesh
+            The given mesh
         points : :class:`str`, optional
             The points to evaluate the field at.
             Either the "centroids" of the mesh cells
@@ -166,10 +166,11 @@ class Field:
         This will store the field in the given mesh under the given name,
         if a meshio or PyVista mesh was given.
 
-        See: https://github.com/nschloe/meshio
-        See: https://github.com/pyvista/pyvista
-
-        See: :any:`__call__`
+        See:
+            - meshio: https://github.com/nschloe/meshio
+            - ogs5py: https://github.com/GeoStat-Framework/ogs5py
+            - PyVista: https://github.com/pyvista/pyvista
+            - Called method: :any:`__call__`
         """
         return generate_on_mesh(self, mesh, points, direction, name, **kwargs)
 
--- a/gstools/field/base.py
+++ b/gstools/field/base.py
@@ -98,7 +98,7 @@ class Field:
             the position tuple, containing main direction and transversal
             directions
         field : :class:`numpy.ndarray` or :any:`None`, optional
-            the field values. Will be all zeros by default.
+            the field values. Will be all zeros if :any:`None` is given.
         mesh_type : :class:`str`, optional
             'structured' / 'unstructured'. Default: 'unstructured'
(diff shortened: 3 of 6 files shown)


LSchueler (changes requested):
I think first of all the `__call__` method and maybe also the `Field` class itself needs more documentation. It is not clear at all what is being "generated" and what the "field values" are.

Now we have `Field.structured`, `Field.unstructured`, `Field.mesh`, and `Field.points`. Wouldn't it be really nice to enhance the `mesh_type` keyword to be able to actually use all these possibilities with only one single method call?

gstools/field/base.py
```diff
@@ -76,32 +79,73 @@ def __init__(
         self._mean = None
         self._normalizer = None
         self._trend = None
+        self._dim = dim if dim is None else int(dim)
```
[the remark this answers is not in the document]
MuellerSeb: Maybe I have just been overcautious. We can also rely on duck-typing ;-)

MuellerSeb:
> I think first of all the `__call__` method and maybe also the `Field` class itself needs more documentation. It is not clear at all what is being "generated" and what the "field values" are.

Maybe just a "Set field with position tuple and values" in the doc header?

> Now we have `Field.structured`, `Field.unstructured`, `Field.mesh`, and `Field.points`. Wouldn't it be really nice to enhance the `mesh_type` keyword to be able to actually use all these possibilities with only one single method call?

This would be an issue for GSTools v2. For now we have to keep these methods. Especially the `Field.mesh` has a lot of settings where we would need to refactor all derived classes to pass `**kwargs` needed for mesh handling. By defining these partial calls to `__call__` as present, we are really flexible since we only rely on the `pos` argument in the `__call__` signature for all these "shortcut" methods. The "mesh_type" itself is only a leftover from the first implementations of GSTools IMO and I think the "(un)structured" options are enough here (See: #61).

Also the planned xarray interface should also be just a new method `Field.xarray` that uses `__call__` internally IMO. Actually I really like this approach to add new interfaces that may need special keywords to control their generation settings (see `mesh`) but internally then uses the simple `__call__` method to generate the desired field. This also makes maintenance  more easy, since we never need to adopt `__call__` or the `mesh_type` property and all child classes smoothly inherit these functionalities.

Here we are again discussing the scope of this class :smile:

LSchueler:
> Maybe just a "Set field with position tuple and values" in the doc header?
Yeah, together with an example of what that is good for is enough I guess. Something like "This gives you access to the plotting routines or the normalization functionality."

LSchueler:
Here we go again ;-)
The extra args for the mesh method is a good point. But Python is known, loved, and hated for its duck-typing. And now you suggest to create a separate field-generation method for all kinds of different input types? Function overloading would be really handy in this case, but I guess Python lacks that feature for a reason.
`structured`, `unstructured`, and `points` (or whatever the best keyword would be) all have the same arguments.
For me the ideal solution would still be to use the different input types to create a `Field` class and then do the field generation with only one method call. But I already tried to implement that and didn't finish, because I don't have enough time for all the refactoring. :'(

Anyways, I think the `points` or `point tuples` or `point list` should be added to the `mesh_type` argument.

MuellerSeb:
> Here we go again ;-)
> The extra args for the mesh method is a good point. But Python is known, loved, and hated for its duck-typing. And now you suggest to create a separate field-generation method for all kinds of different input types? Function overloading would be really handy in this case, but I guess Python lacks that feature for a reason.
> `structured`, `unstructured`, and `points` (or whatever the best keyword would be) all have the same arguments.
> For me the ideal solution would still be to use the different input types to create a `Field` class and then do the field generation with only one method call. But I already tried to implement that and didn't finish, because I don't have enough time for all the refactoring. :'(
> 
> Anyways, I think the `points` or `point tuples` or `point list` should be added to the `mesh_type` argument.

Then I would vote for kicking `points` out again, since all it does is transposing the `pos` tuple. When adding a new `mesh_type` possibility, we would also need to refactor all downstream routines (exporting, plotting, converting). We added `structured` and `unstructured` only for convenience back in the days.

LSchueler:
We actually added the `structured` and `unstructured` methods for convenience, not the `mesh_type` keywords. However, we shouldn't use history as arguments for the future development ;-)

I have to agree that it would take a lot of careful considerations, if we introduce a new `mesh_type` keyword. Although I still do prefer it that way. But if we now introduce a method `points` and later maybe remove it or add `mesh_type` keywords, that could get pretty confusing. Maybe we should indeed drop the `points` method for the moment.

gstools/field/base.py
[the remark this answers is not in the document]
MuellerSeb: I always think about some string input stuff, where people read configurations from a file or config-string, and they pass `"1.0"` or something similar.

MuellerSeb:
Ok, then I am fine with dropping `points` again. ;-) I think passing a `points.T` for `pos` is also fine.

MuellerSeb:
@LSchueler removed the `points` method and added an example. LGTM!

CHANGELOG.md
```diff
@@ -2,6 +2,16 @@
 
 All notable changes to **GSTools** will be documented in this file.
 
+## [1.3.1] - Pure Pink - 2021-?
```
LSchueler, line 5: Better not forget the month later ;-)

examples/00_misc/05_standalone_field.py
```diff
@@ -0,0 +1,27 @@
...
+random values and plot them.
+"""
+import numpy as np
+import gstools as gs
+
+x1 = np.random.RandomState(19970221).rand(10000) * 100.0
+x2 = np.random.RandomState(20011012).rand(10000) * 100.0
+x3 = np.random.RandomState(20210530).rand(10000) * 100.0
+x4 = np.random.RandomState(20210531).rand(10000) * 100.0
+values = np.random.RandomState(2021).rand(10000) * 100.0
```
LSchueler, line 17: I think it would be nice to utilise our MasterRNG here to generate the different RNG-streams, instead of instantiating 5 single ones on your own. On the other hand, this clutters the example a bit. What do you think?

examples/00_misc/05_standalone_field.py
```diff
@@ -0,0 +1,27 @@
...
+
+###############################################################################
+# Only thing needed to instantiate the Field is the dimension.
+#
+# Afterwards we can call the instance like all other Fields
+# (:any:`SRF`, :any:`Krige` or :any:`CondSRF`), but with an additional field.
+
+plotter = gs.field.Field(dim=4)
```
LSchueler, line 25: Maybe you could import gstools here instead of at the beginning of the script to emphasize that gstools is not needed to generate the field? - But this is a rather uncommon practice in Python, which could potentially confuse people...

gstools/field/base.py
```diff
@@ -97,10 +97,10 @@ def __call__(
         pos : :class:`list`
             the position tuple, containing main direction and transversal
             directions
-        field : :class:`numpy.ndarray`
-            the field values.
-        mesh_type : :class:`str`
-            'structured' / 'unstructured'
+        field : :class:`numpy.ndarray` or :any:`None`, optional
+            the field values. Will be all zeros by default.
```
LSchueler, line 101: Better phrase "will be np.zeros_like(pos) if None given"?

LSchueler (changes requested):
I think it's good to have the example.
And I think we're nearly there ;-)

gstools/field/base.py
MuellerSeb: Problem is, that `np.zeros_like(pos)` is not true. Neither for unstructured meshes (`np.zeros_like(pos[0])`, can even be different in 1D) nor structured meshes (`np.zeros(tuple(map(np.size, pos)))`). So I think *"Will be all zeros"* is quite descriptive at this point.

examples/00_misc/05_standalone_field.py
MuellerSeb: I would vote against importing not at the top of the script. There is also a description stating that these values are not generated with gstools.

examples/00_misc/05_standalone_field.py
MuellerSeb: Why not. What do you think about this:
```python
rng = gs.random.RNG(19970221)
x1 = rng.random.rand(10000) * 100.0
x2 = rng.random.rand(10000) * 100.0
x3 = rng.random.rand(10000) * 100.0
x4 = rng.random.rand(10000) * 100.0
values = rng.random.rand(10000) * 100.0
```
?

CHANGELOG.md
MuellerSeb: Jupp! Got that on the radar!

examples/00_misc/05_standalone_field.py
MuellerSeb: But with using `RNG` of gstools (as suggested in the comment above), they are again generated with gstools :joy:

examples/00_misc/05_standalone_field.py
LSchueler: That looks a lot nicer!

LSchueler: Hmm, now that I thought about it again, a single RNG stream for all the values shouldn't be a problem at all. It's only a problem, if you take parts of the generated numbers and separate them, like
```python
for t in time:
    x += rng()
    y += rng()
````
There you should be using
```python
for t in time:
    x += rng1()
    y += rng2()
````
But in your example you take all the subsequent generated numbers for one dimension and then for the next.
So, you could use

```python
rng = np.random.RandomState(19970221)
x1 = rng.random.rand(10000) * 100.0
x2 = rng.random.rand(10000) * 100.0
x3 = rng.random.rand(10000) * 100.0
x4 = rng.random.rand(10000) * 100.0
values = rng.random.rand(10000) * 100.0
```

examples/00_misc/05_standalone_field.py
LSchueler: > But with using `RNG` of gstools (as suggested in the comment above), they are again generated with gstools joy

Yeah I know... but see my other comment.

gstools/field/base.py
LSchueler: My main point was actually the "by default". I'm just not sure if that could be confusing if handing over your own field.
And yeah, my `np.zeros_like(pos)` is some kind of pseudo code.

MuellerSeb: Ok, so I will replace the "by default" with "if None is given".

examples/00_misc/05_standalone_field.py
MuellerSeb: OK, so to not confuse users, I will keep it at the top.

examples/00_misc/05_standalone_field.py
MuellerSeb: Will use the latter one. with one instance of `RandomState`.

LSchueler (approved):
We finally made it! :+1:

MuellerSeb:
> We finally made it! +1

Yeah! Will merge when CI is successful. :tada:

PR #169, opened 2021-04-28
README: add social contact links

MuellerSeb:
Add links to:
- GH Discussions
- Slack channel in Software Underground
- Gitter Chat


LSchueler (approved):
:+1:
Should we add our preferred channel? - I mean I guess the order is an implicit hint ;-)

MuellerSeb:
> +1
> Should we add our preferred channel? - I mean I guess the order is an implicit hint ;-)

I choose that order, sine IMHO this is our current ranking. ;-)

Issue #179, opened 2021-06-09
BUG: vario_estimate alters field


MuellerSeb:
Problem: `gs.vario_estimate` could alter the input field, when given as a numpy array with floats along with a trend function:
```python
import numpy as np
import gstools as gs

trend = lambda a, b: a
x = y = t = np.array([1.0, 2.0, 3.0, 4.0])
gs.vario_estimate((x, y), t, trend=trend)
print(t)
```
results in
```bash
[0. 0. 0. 0.]
```
Although it should be `[1. 2. 3. 4.]`

To solve this, we need to set `copy=True` here:
https://github.com/GeoStat-Framework/GSTools/blob/f91d3b12ef3f278c85eac84c7a9a1bb3c7e7bf55/gstools/variogram/variogram.py#L246

PR #180, opened 2021-06-09
[fix] vario_estimate: prevent altering the input field

MuellerSeb:
closes: #179


LSchueler (commented):
Nice catch!

Maybe we should systematically add some unit tests to check for such things, but I guess that is not the scope of this PR.

LSchueler (approved):

PR #204, opened 2021-08-20
Cython cleanup

LSchueler:
I cleaned up some of the older Cython code and I could actually reuse two functions and delete duplicate code.

### DIFF
--- a/gstools/field/summator.pyx
+++ b/gstools/field/summator.pyx
@@ -12,31 +12,26 @@ from libc.math cimport sin, cos
 cimport numpy as np
 
 
-DTYPE = np.double
-ctypedef np.double_t DTYPE_t
-
-
 def summate(
-    const double[:,:] cov_samples,
+    const double[:, :] cov_samples,
     const double[:] z_1,
     const double[:] z_2,
-    const double[:,:] pos
+    const double[:, :] pos
     ):
-    cdef int i, j, d, X_len, N
+    cdef int i, j, d
     cdef double phase
-    cdef int dim
-    dim = pos.shape[0]
+    cdef int dim = pos.shape[0]
 
-    X_len = pos.shape[1]
-    N = cov_samples.shape[1]
+    cdef int X_len = pos.shape[1]
+    cdef int N = cov_samples.shape[1]
 
-    cdef double[:] summed_modes = np.zeros(X_len, dtype=DTYPE)
+    cdef double[:] summed_modes = np.zeros(X_len)
 
     for i in prange(X_len, nogil=True):
         for j in range(N):
             phase = 0.
             for d in range(dim):
-                phase += cov_samples[d,j] * pos[d,i]
+                phase += cov_samples[d, j] * pos[d, i]
             summed_modes[i] += z_1[j] * cos(phase) + z_2[j] * sin(phase)
 
     return np.asarray(summed_modes)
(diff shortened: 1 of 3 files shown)


gstools/field/summator.pyx
```diff
@@ -12,31 +12,26 @@ from libc.math cimport sin, cos
...
-    cdef int dim
-    dim = pos.shape[0]
+    cdef int dim = pos.shape[0]
 
-    X_len = pos.shape[1]
-    N = cov_samples.shape[1]
+    cdef int X_len = pos.shape[1]
+    cdef int N = cov_samples.shape[1]
 
-    cdef double[:] summed_modes = np.zeros(X_len, dtype=DTYPE)
```
MuellerSeb, line 33: So is using `dtype` really redundant here? It seems I know to little about cython-numpy integration.

gstools/field/summator.pyx
```diff
@@ -12,31 +12,26 @@ from libc.math cimport sin, cos
...
+    cdef double[:] summed_modes = np.zeros(X_len)
 
     for i in prange(X_len, nogil=True):
         for j in range(N):
             phase = 0.
             for d in range(dim):
-                phase += cov_samples[d,j] * pos[d,i]
+                phase += cov_samples[d, j] * pos[d, i]
```
MuellerSeb, line 34: Unfortunately `black` is still not able to deal with `*.pyx` files: https://github.com/psf/black/issues/359

gstools/variogram/estimator.pyx
```diff
@@ -106,21 +105,21 @@ cdef inline double estimator_cressie(const double f_diff) nogil:
 ctypedef double (*_estimator_func)(const double) nogil
 
 cdef inline void normalization_matheron(
-    vector[double]& variogram,
-    vector[long]& counts
+    double[:] variogram,
+    long[:] counts,
```
MuellerSeb, line 109: I thought `vector` is faster? And is this still using the references? (formerly marked with `&`) Or is this done automatically by inlining the function? Again: I could use some training in cython.

gstools/variogram/estimator.pyx
```diff
@@ -129,38 +128,30 @@ cdef inline void normalization_cressie(
...
     cdef long cnt
     for d in range(variogram.shape[0]):
-        for i in range(variogram.shape[1]):
-            # avoid division by zero
-            cnt = max(counts[d, i], 1)
-            variogram[d, i] = (
-                0.5 * (1./cnt * variogram[d, i])**4 /
-                (0.457 + 0.494 / cnt + 0.045 / cnt**2)
-            )
+        normalization_cressie(variogram[d, :], counts[d, :])
```
MuellerSeb, line 150: Good catch!

gstools/variogram/estimator.pyx
```diff
@@ -129,38 +128,30 @@ cdef inline void normalization_cressie(
...
-    long[:,:]& counts
+    double[:,:] variogram,
+    long[:,:] counts,
 ):
     cdef int d, i
     for d in range(variogram.shape[0]):
-        for i in range(variogram.shape[1]):
-            # avoid division by zero
-            variogram[d, i] /= (2. * max(counts[d, i], 1))
+        normalization_matheron(variogram[d, :], counts[d, :])
```
MuellerSeb, line 141: Good catch!

gstools/variogram/estimator.pyx
```diff
@@ -345,8 +332,8 @@ def ma_structured(
     cdef int j_max = f.shape[1]
     cdef int k_max = i_max + 1
 
-    cdef vector[double] variogram = vector[double](k_max, 0.0)
-    cdef vector[long] counts = vector[long](k_max, 0)
+    cdef double[:] variogram = np.zeros(k_max)
+    cdef long[:] counts = np.zeros(k_max, dtype=long)
```
MuellerSeb, line 336: Ah! So you need to specify dtype if it's not `double` right? I hope numpy doesn't change its default behavior to soon.

MuellerSeb (commented):
Nice cleanup! One thing following "explicit is better than implicit": Wouldn't it be better to always use `dtype` when initializing numpy arrays in cython to not run into possible errors in the future? Maybe I am a bit overcautious here...

gstools/variogram/estimator.pyx
LSchueler: It is faster, but I think only during the initialization of the variable. It does not call any Python code, in contrast to the `cdef double[:, :] variogram = np.zeros(...)`. But I'm very sure that that is negligible and I think the code is a lot cleaner when only using Cython's memoryviews and not mixed up with C++ data types.

LSchueler:
I was actually also very unsure if the `dtype` stuff should go or not. Sure, explicit is better than implicit, but if Python would stay true to that, it shouldn't allow default arguments.
And I cannot imagine numpy to change its default `dtype` from 64 bit floats to anything else. But as I said, I'm pretty unsure about removing that. If you think we should stick to Python's (sometimes hypocritical ;-) ) zen code, then I'd be happy to change the code again.

LSchueler:
@MuellerSeb Have you made up your mind about the `dtype`?

I think I will save the topic of contiguous arrays for another PR.

gstools/variogram/estimator.pyx
MuellerSeb: Nice.

MuellerSeb:
@LSchueler : I'd like to see an explicit `dtype` setting in the `np.zeros` call. And one other thing that came to my mind: Couldn't we remove the line `# distutils: language = c++` in `extimator.pyx` now since we don't use `vector`? After that I am happy to merge.

MuellerSeb:
Also this line can be removed I guess:
https://github.com/GeoStat-Framework/GSTools/blob/3e4b6c36bd89fe3a8ca11fbde60f67fb1be0a5d2/setup.py#L146

LSchueler:
I think we should keep the C++ flag. It doesn't hurt us, the same compiler is used anyway and this way, in case one day we want to use a C++ feature, we don't have to dig through the Cython docs again.

I added the dtypes again to the numpy functions.

PR #215, opened 2021-10-27
Add optional dependency GSTools-Core

LSchueler:
I removed all Cython code and replaced it with the GSTools-Core PyPI package v0.1.2.

The Github actions still need to be updated, as the distribution of GSTools has become muuch simpler.


LSchueler:
Nearly all of the tests done with during the `sdist` [job](https://github.com/GeoStat-Framework/GSTools/runs/4024772715?check_suite_focus=true) rely on the Rust implementations. But as you have discovered recently, we still don't cover everything. So, there is some work to do!
The other jobs are only failing due to some import linter problem and that we don't need to build wheels anymore, as GSTools is now a pure Python module.

adamreichold:
> Nearly all of the tests done with during the sdist job rely on the Rust implementations. But as you have discovered recently, we still don't cover everything. So, there is some work to do!

I was pondering something more automated, e.g. using the existing implementation as a model-based test using a coverage-guided fuzzer that would automatically create test cases with the definition of the test cases basically being to yield the same results as the existing implementation. I am just note sure how to integrate the existing implementation, i.e. on a source-level or maybe by just tapping into a fixed binary distribution.

MuellerSeb:
After a nice conversation with @LSchueler we decided:
- use `gstools_core` as an optional drop-in replacement and add a switch to control the used backend
  ```python
  try:
      import gstools_core

  except ImportError:
      USE_RUST = False
  finally:
      USE_RUST = True
      del gstools_core
  ```
- add `gstools_core` as optional dep. to `setup.cfg` to install it with `pip install gstools[rust]`
- solve #216

adamreichold:
> use gstools_core as an optional drop-in replacement and add a switch to control the used backend 

Shouldn't this be
```python
try:
    import gstools_core
    USE_RUST = True
except ImportError:
    del gstools_core
    USE_RUST = False
```
i.e. only if the import succeeds, `USE_RUST == True`, not unconditionally as the finally block is executed?

MuellerSeb:
I meant `else` ... Sorry

LSchueler:
The Cython code is back in again and GSTools-Core is going to be used if the package is installed and `gstools.config.USE_RUST` is not manually set to `False`
I also refactored the Cython code according to issue #216 , which would close it.

docs/source/index.rst
```diff
@@ -89,6 +91,31 @@ For the development version, you can do almost the same:
...
+
+    pip install gstools[rust]
+
+or by manually installing the package
+
+.. code-block:: none
+
+    pip install gstools-core
+
+GSTools-Core will automatically use all your cores in parallel, without having
```
adamreichold, line 116: Since one sometimes wants to restrict parallelism (I am looking at you unisolated HPC job scheduling systems... :roll_eyes:), it might be nice to mention that the environment variable `RAYON_NUM_THREADS` controls the parallelism for this backend.

docs/source/index.rst
```diff
@@ -387,6 +414,7 @@ Requirements
 Optional
 --------
 
+- `GSTools-Core >= 0.1.0 <https://github.com/GeoStat-Framework/GSTools-Core>`_
```
adamreichold, line 417: I think it needs at least 0.2.0 because this when the `dim` argument was removed.

gstools/variogram/variogram.py
```diff
@@ -20,12 +22,29 @@
...
-from gstools.variogram.estimator import (
-    directional,
-    ma_structured,
-    structured,
-    unstructured,
-)
+
+if config.USE_RUST:
+    # pylint: disable=E0401
+    from gstools_core import (
```
adamreichold, line 28: Shouldn't it be possible [to merge those](https://stackoverflow.com/questions/29865421/how-can-i-import-multiple-items-from-a-module-and-rename-them-in-python) or this is a formatter issue?

setup.cfg
```diff
@@ -66,6 +66,8 @@ doc =
 plotting =
     matplotlib>=3,<4
     pyvista>=0.29,<1
+rust =
+    gstools_core>=0.1.2,<1
```
adamreichold, line 70: As above, I think `>=0.2.0,<1` is technically required.

docs/source/index.rst
LSchueler: That's an excellent point! I'll add it to the readme right away.

gstools/variogram/variogram.py
LSchueler: Yeah... that was `isort`'s idea. It's about as ugly as it gets, right?! :-D And `black` and `isort` can't even agree on where to put the `# pragma: no cover`.

docs/source/index.rst
LSchueler: True!

setup.cfg
LSchueler: Even more important ;-)

LSchueler:
For the moment, we don't provide GSTools-Core via conda, which makes it a bit awkward to use it, if GSTools is installed via conda. But I think that's a problem of GSTools-Core and as the Rust implementations are labeled as experimental anyway, I think this is ready to be merged.

docs/source/index.rst
```diff
@@ -89,6 +91,33 @@ For the development version, you can do almost the same:
...
+or by manually installing the package
+
+.. code-block:: none
+
+    pip install gstools-core
+
+GSTools-Core will automatically use all your cores in parallel, without having
+to use OpenMP or a local C compiler.
+In case you want to restrict the number of threads used, you can set the
+environment variable ``RAYON_NUM_THREADS`` to the desired amount.
```
MuellerSeb, line 119: What about a little function in `gstools.config` to set this environment variable?

gstools/config.py
```diff
@@ -0,0 +1,14 @@
+# -*- coding: utf-8 -*-
+"""
+GStools subpackage providing global variables.
+
+.. currentmodule:: gstools.config
+
+"""
+# pylint: disable=W0611
+try:
```
MuellerSeb, line 9: What about setting `# pragma: no cover` behind `try:`?

gstools/field/generator.py
```diff
@@ -10,16 +10,23 @@
...
 from gstools.covmodel.base import CovModel
-from gstools.field.summator import summate, summate_incompr
 from gstools.random.rng import RNG
 
+if config.USE_RUST:
+    # pylint: disable=E0401
+    from gstools_core import summate, summate_incompr  # pragma: no cover
+else:
+    # pylint: disable=C0412
+    from gstools.field.summator import summate, summate_incompr
```
MuellerSeb, line 28: What about setting `# pragma: no cover` behind `if config.USE_RUST:`?
Also the `# pylint: disable=C0412` is redundant (see line 13) and we could also just add `E0401` in line 13.

gstools/krige/base.py
```diff
@@ -9,23 +9,32 @@
...
+    # pylint: disable=E0401
+    from gstools_core import (  # pragma: no cover
+        calc_field_krige,
+        calc_field_krige_and_variance,
+    )
+else:
+    from gstools.krige.krigesum import (
+        calc_field_krige,
+        calc_field_krige_and_variance,
+    )
```
MuellerSeb, line 36: What about setting `# pragma: no cover` behind `if config.USE_RUST:`?
We could also just add `E0401` in line 13.

gstools/variogram/variogram.py
MuellerSeb: What about setting `# pragma: no cover` behind `if config.USE_RUST:` once?
Also the `# pylint: disable=C0412` is redundant (see line 13) and we could also just add `E0401` in line 13.

MuellerSeb (changes requested):
This is really cool! I got some ideas for improvement, maybe you like them. And what about a routine to set the env-var for the number of cores for the rust parallel runs in `gstools.config`?

docs/source/index.rst
adamreichold: One argument against that is that this variable will affect all Rust software using Rayon, not just GSTools-Core's backend.

If we do provide a programmatic interface, we should probably make the Core initialize a dedicated thread pool and make this function affect only this thread pool?

gstools/variogram/variogram.py
LSchueler: Sorry, I think I was a bit annoyed yesterday by the combination of a false positive coming from pylint and then the isort and black peskiness. I imnproved the situation, but it seems that isort does not like grouped renaming of imports.
I don't want to disable the pylint error `E0401` globally, as import errors can be quite a severe problem.

docs/source/index.rst
LSchueler: @adamreichold raises a fair point there. But that env. variable would only exist during the shell session from which GSTools is run (I don't have enough experience with Windows to know how it handles env. variables).
Did I understand it correctly, that the number of threads used by Rayon can only be set once at the start and doing it a second time results in an error? I'm not if this is worth the effort.
Another alternative would be to capture the env. variable at the start of GSTools and set it back to that value when exiting, but that doesn't sound very elegant.
I personally would just keep the hint in the readme for the moment. But if you have stronger opinions than I do, I'd be happy to implement your preferred solution.

LSchueler:
I think an import error (E0401) should only be disabled locally.

docs/source/index.rst
adamreichold: > Did I understand it correctly, that the number of threads used by Rayon can only be set once at the start and doing it a second time results in an error? I'm not if this is worth the effort.

Setting the environment variable will affect Rayon's so-called [global](https://docs.rs/rayon/1.5.1/rayon/struct.ThreadPoolBuilder.html#method.build_global) thread pool which is initialized once, either on-demand when the first Rayon task is spawn or explicitly be the application. Hence, changing the environment variable after a first computation using Rayon was run has no effect.

This why I suggested making the GSTools-Core module store an `Option<ThreadPool>` which would be affected (meaning re-initialized using a given number of threads) by the function from `gstools.config`. If the option is set, that thread pool is used via [`ThreadPool::install`](https://docs.rs/rayon/1.5.1/rayon/struct.ThreadPool.html#method.install), otherwise GSTools-Core falls back to the global thread pool.

> I personally would just keep the hint in the readme for the moment.

Personally, I would say that online control over parallelism is a somewhat orthogonal issue and would best be served via a follow-up PR / issue.

MuellerSeb (approved):
LGTM

PR #227, opened 2021-12-15
Field: make Field.mesh compatible with meshio v5.1+

### LINKED ISSUE #226: Incompatible with MeshIO v5.1
Executing the [mesh ensemble example](https://geostat-framework.readthedocs.io/projects/gstools/en/stable/examples/01_random_field/05_mesh_ensemble.html#sphx-glr-examples-01-random-field-05-mesh-ensemble-py) gives the following Error:
```bash
TypeError: 'CellBlock' object is not subscriptable
```
This is due to the latest update of meshio: https://github.com/nschloe/meshio/releases/tag/v5.1.0
There a change was introduced on how to use CellBlocks: https://github.com/nschloe/meshio/blob/main/CHANGELOG.md#v510-dec-11-2021

MuellerSeb:
Closes #226


LSchueler (changes requested):
Thanks for this bugfix. :+1:

But why do you think we should continue supporting older meshio versions? If someone wants to update `GSTools`, `meshio` can also be updated. I fear that this could make it much more difficult to maintain or own code, with mutiple branching dependencies.
And I just checked, `meshio` does not provide bugfixes for older major versions.

MuellerSeb:
Guess you are right. So I will bump the minimum version of meshio to 5.1 and remove the upper restriction to not make it a blocker in the future.

examples/04_vector_field/01_3d_vector_field.py
```diff
@@ -16,8 +16,8 @@
 
 ###############################################################################
 # create a uniform grid with PyVista
-dim, spacing, origin = (40, 30, 10), (1, 1, 1), (-10, 0, 0)
-mesh = pv.UniformGrid(dim, spacing, origin)
+dims, spacing, origin = (40, 30, 10), (1, 1, 1), (-10, 0, 0)
+mesh = pv.UniformGrid(dims=dims, spacing=spacing, origin=origin)
```
LSchueler, line 20: Why so ugly? Are you afraid that the ordering of the arguments might change? - If that is the case, I would argue that it would be similar likely that the keywords themselves could change.

setup.cfg
```diff
@@ -47,12 +47,12 @@ project_urls =
...
-    meshio>=4.0.3,<6
-    numpy>=1.14.5,<2
-    pyevtk>=1.1.1,<2
-    scipy>=1.1.0,<2
+    emcee>=3.0.0
+    hankel>=1.0.0
+    meshio>=4.0.0
+    numpy>=1.14.5
+    pyevtk>=1.1.1
+    scipy>=1.1.0
```
LSchueler, line 55: I think this is a really good thing :+1:

LSchueler (changes requested):
What happened to the "... I will bump the minimum version of meshio to 5.1..."?

examples/04_vector_field/01_3d_vector_field.py
MuellerSeb: This is why ReadTheDocs failed. I just fixed it to work. And yes, this was the given solution from the docs of pyvista (from v0.33 on):
> Changed in version 0.33.0: First argument must now be either a path or vtk.vtkImageData. Use keyword arguments to specify the dimensions, spacing, and origin of the uniform grid.

See: https://docs.pyvista.org/api/core/_autosummary/pyvista.UniformGrid.html?highlight=uniformgrid#pyvista.UniformGrid

MuellerSeb:
> What happened to the "... I will bump the minimum version of meshio to 5.1..."?

?! I did this before, but to support py36 as discussed, I needed to support meshio 4 since meshio 5+ only supports py37+. That is why I reverted this as we agreed on.

LSchueler (approved):
Okay... I need a coffee, some fresh air and shouldn't work as much ;-)

MuellerSeb:
Trying to nail it down. This looks strange in the log for `ubuntu i686`:
```
2022-01-12T21:36:56.7636311Z Collecting numpy>=1.14.5
2022-01-12T21:36:56.7637269Z   Downloading numpy-1.22.0.zip (11.3 MB)
2022-01-12T21:36:56.7638159Z   Installing build dependencies: started
2022-01-12T21:36:56.7639410Z   Installing build dependencies: finished with status 'done'
2022-01-12T21:36:56.7640459Z   Getting requirements to build wheel: started
2022-01-12T21:36:56.7641701Z   Getting requirements to build wheel: finished with status 'done'
2022-01-12T21:36:56.7642735Z     Preparing wheel metadata: started
2022-01-12T21:36:56.7644306Z     Preparing wheel metadata: finished with status 'done'
```
last working version looked like this:
```
2021-12-15T13:54:46.9763629Z Collecting numpy<2,>=1.14.5
2021-12-15T13:54:46.9765160Z   Downloading numpy-1.21.4-cp38-cp38-manylinux_2_12_i686.manylinux2010_i686.whl (13.7 MB)
```
It looks like numpy is now build from source whereas a wheels was taken last time. For py37 it takes a wheel:
```
2022-01-12T21:29:50.6808690Z Collecting numpy>=1.14.5
2022-01-12T21:29:50.6809449Z   Downloading numpy-1.21.5-cp37-cp37m-manylinux_2_12_i686.manylinux2010_i686.whl (13.7 MB)
```
Maybe we can pin the numpy version for testing to `<1.22`.

PR #229, opened 2022-01-15
CI: PIP_PREFER_BINARY=1 to prevent building deps

MuellerSeb:
A better solution than capping numpy.
See here: https://github.com/pypa/cibuildwheel/issues/993


LSchueler (approved):
Thanks for taking care of this!! Approved!

PR #241, opened 2022-05-27
Build: update setuptools config and CI

MuellerSeb:
Changes:
- drop Python 3.6 support (setuptools>60 needs py>3.7)
- move `setup.cfg` content to `pyproject.toml` ([PEP 621](https://peps.python.org/pep-0621/))
- update cibuildwheel to 2.6 (requirement for `pyproject.toml` config with setuptools)
- move to `src/` based package structure (better for testing, building and structure)
- use [extension-helpers](https://pypi.org/project/extension-helpers/) for openmp support in `setup.py`


MuellerSeb:
OpenMP support was never available in CI for macOS. There is also an open PR for clang/openmp support on macOS here: https://github.com/astropy/extension-helpers/pull/42

We could wait for this PR to be merged and then we could install openmp with brew for macOS:
```yml
      - name: macos openmp install
        uses: mstksg/get-package@v1
        if: matrix.os == 'macos-latest'
        with:
          brew: libomp
```

pyproject.toml
```diff
@@ -1,15 +1,94 @@
...
+    "Intended Audience :: Education",
+    "License :: OSI Approved :: GNU Lesser General Public License v3 (LGPLv3)",
+    "Natural Language :: English",
+    "Operating System :: Unix",
+    "Operating System :: Microsoft",
+    "Operating System :: MacOS",
+    "Programming Language :: Python",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.6",
```
LSchueler, line 41: Drop this line

LSchueler (changes requested):
That's a nice improvement in the setup! A small fix in the toml-file and you're ready to merge this :+1:

MuellerSeb:
When this is implemented, we can even drop the `MANIFEST.in`: https://github.com/pypa/setuptools/issues/3341

PR #243, opened 2022-06-20
Add: (Exponential-)Integral model

MuellerSeb:
- adding a new model derived from the Exponential Integral named `Integral`
- fix approximation of Matern spectrum for big `nu`

### DIFF
--- a/examples/02_cov_model/README.rst
+++ b/examples/02_cov_model/README.rst
@@ -62,8 +62,7 @@ The following standard covariance models are provided by GSTools
     Gaussian
     Exponential
     Matern
-    ExpInt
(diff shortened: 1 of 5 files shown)


LSchueler (commented):
Gimme more cov. models! :-) Nice!
I think it would be better, if you explicitly test the `ExpInt` model in `test_covmodel.py`.

I was never sure how much error checking we should include in the Cython-code. But as this will hopefully one day be legacy code, I think your fix is fine.

MuellerSeb:
Rebase after #255 worked fine.

LSchueler (approved):
Great, LGTM!

Issue #246, opened 2022-08-04
`scipy.linalg.pinv2` not available in scipy `1.9`


mmaelicke:
Hey Sebastian,
I am running some docker containers with always the latest gstools version in it. Now, the import of gstools fails.
Here is the relevant part of the traceback:

```
File "/usr/local/lib/python3.10/site-packages/skgstat/interfaces/gstools.py", line 151, in skgstat_to_krige
    import gstools as gs
File "/usr/local/lib/python3.10/site-packages/gstools/__init__.py", line 129, in <module>
    from gstools import (
File "/usr/local/lib/python3.10/site-packages/gstools/field/__init__.py", line 33, in <module>
    from gstools.field.cond_srf import CondSRF
File "/usr/local/lib/python3.10/site-packages/gstools/field/cond_srf.py", line 17, in <module>
    from gstools.krige import Krige
File "/usr/local/lib/python3.10/site-packages/gstools/krige/__init__.py", line 20, in <module>
    from gstools.krige.base import Krige
File "/usr/local/lib/python3.10/site-packages/gstools/krige/base.py", line 38, in <module>
    P_INV = {"pinv": spl.pinv, "pinv2": spl.pinv2, "pinvh": spl.pinvh}
```

I could verify, that `scipy==1.9.0` in fact has no `scipy.linalg.pinv2` method, while `scipy==1.8.0` imports it just fine. The whole practice was run in a Python 3.10 environment.
I solved the issue by installing scipy==1.8.0 first and then gstools. I am also not sure if the `pinv2` got moved or removed and if therefore any action on gstools-side is necessary.

mmaelicke:
https://github.com/scipy/scipy/issues/15737  - Seems like `pinv2` should be deprecated and/or removed without a deprecation warning?

LSchueler:
Hi Mirko,
thanks for reporting this! It should be resolved with PR #247 .

MuellerSeb:
Also see: https://github.com/GeoStat-Framework/PyKrige/issues/236

PR #247, opened 2022-08-04
Remove deprecated scipy method pinv2

LSchueler:
With Scipy v1.9 GSTools fails to import `pinv2`, as reported in #246 .

According to the [PR](https://github.com/scipy/scipy/pull/13831) both methods `pinv` and `pinv2` have been using the same algorithms for a long time. Thus, I simply deleted the option `pinv2`.

### DIFF
--- a/src/gstools/krige/base.py
+++ b/src/gstools/krige/base.py
@@ -35,7 +35,7 @@ else:
 __all__ = ["Krige"]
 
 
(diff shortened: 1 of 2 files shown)


LSchueler:
The failed tests are not a problem of this PR and they are fixed with PR #248 .

LSchueler:
Now, all CI runs fail due to the updated scipy-package. I'll just merge this right now.

PR #248, opened 2022-08-04
Change float comparison in tests

LSchueler:
Some tests are failing in PR #247 due to very small discrepancies (~10e-16) in floating point comparisons. An approximate comparison is the better test.


MuellerSeb (approved):
LGTM

Issue #235, opened 2022-04-20
I14Y: latlon support for PyKrige interface


MuellerSeb:
Ordinary Kriging in PyKrige supports geographical coordinates, but passing a GSTools Covmodel with `latlon=True` results in an error:
https://github.com/GeoStat-Framework/PyKrige/issues/217

On the GSTools side, we need to update `pykrige_vario` to be the yadrenko variogram with input in degrees.

PR #254, opened 2022-08-10
pykrige: fix passed variogram in case of geo-coord

MuellerSeb:
PyKrige was not able to use geographical coordinates with a GSTools CovModel. This is the GSTools part of the fix.

See: https://github.com/GeoStat-Framework/PyKrige/issues/217 and https://github.com/GeoStat-Framework/GSTools/issues/235


LSchueler (approved):
This looks like a solid solution :-) :+1:

PR #255, opened 2022-08-10
Cherry picks minor fixes for v1.4

MuellerSeb:
Cherry picks from #243 to be included in v1.4:
- add bounds checks for optional arguments when resetting by class attribute
- Cython: solve -Wsometimes-uninitialized warning
- minor coverage improvements

### DIFF
--- a/src/gstools/__init__.py
+++ b/src/gstools/__init__.py
@@ -180,7 +180,7 @@ from gstools.variogram import (
 
 try:
     from gstools._version import __version__
-except ModuleNotFoundError:  # pragma: nocover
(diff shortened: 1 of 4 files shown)


LSchueler (commented):
These are all sensible changes, but I fear that this will lead directly into merge hell. How about merging PR #243 and remove the cherry picking from this PR?

MuellerSeb:
> These are all sensible changes, but I fear that this will lead directly into merge hell. How about merging PR #243 and remove the cherry picking from this PR?

But merging this and rebasing #243 should also work, or? I don't want to include #243 in 1.4 just now, that is why I cherry picked the fixes from there.

PR #259, opened 2022-08-16
Generator: add abstract __init__

MuellerSeb:
The abstract `__init__` was missing in the abstract generator base class.


MuellerSeb:
Needed to ignore `W0231` in pylint (super-init-not-called), since abstract `__init__` routines will only be excluded from that rule with the next pylint release (2.15).

See: https://github.com/PyCQA/pylint/issues/3975

PR #269, opened 2022-10-19
seed: use Ellipsis to indicate keeping current seed

MuellerSeb:
We used `np.nan` to indicate to keep the current seed in the RNG, which is a bit of a hack IMHO.

Python provides a special built in constant to be used as placeholder: [Ellipsis](https://docs.python.org/3/library/constants.html#Ellipsis)

In my opinion it is more intuitive to state `seed=...` to keep the seed in contrast to `seed=np.nan` since we can't use `None` which is a special value for seeds. Since Python provides a dedicated constant to be used as a placeholder, I would use it here.


LSchueler:
Hmm, I don't have very strong feelings about this one. I don't see `np.nan` to be a hack, just a value, we decided on. I have actually never seen Python ellipses outside of slicing operations. Maybe I always ignored ellipses outside of slices, but are you not concerned that people could get confused when they see `...` in an argument list?

MuellerSeb:
Maybe you are right. np.nan still makes sense, I was just happy about learning, that Ellipsis is used as a placeholder as described here for example: https://realpython.com/python-ellipsis/#in-short-use-the-ellipsis-as-a-placeholder-in-python

We can also keep the current behavior.

LSchueler:
Woah! You can even replace `pass` with `...`?!?
:-) I mean, it is pretty fancy and if it would be more common, I think I would definitely be for dropping `np.nan` as the default argument.
I'll let you decide what to do with it.

LSchueler (approved):
It's up to you ;-)

MuellerSeb:
Will keep it as a draft for now. And yes, Ellipsis are fancy :-)

PR #273, opened 2022-10-29
Field.plot: solve long-standing mpl slider bug

MuellerSeb:
We had a long standing bug in the used matplotlib sliders in the nD viewers as seen here:
https://geostat-framework.readthedocs.io/projects/gstools/en/v1.4.0/examples/01_random_field/07_higher_dimensions.html


LSchueler (approved):
That looks easy :-), LGTM!

PR #283, opened 2022-11-11
Remove unnecessary coding, use a few f-strings

mwtoews:
These edits were automated using:

    pyupgrade --py37-plus `find . -name "*.py*"`

and does:

- Remove unnecessary `# -*- coding: utf-8 -*-` comment, since [PEP 3120](https://peps.python.org/pep-3120/) declared UTF-8 as the default source encoding since Python 3.0
- Convert a few f-strings
- Add missing `r` prefix for a raw string with `\` used for math markup

### DIFF
--- a/src/gstools/variogram/estimator.pyx
+++ b/src/gstools/variogram/estimator.pyx
@@ -1,6 +1,5 @@
 #cython: language_level=3, boundscheck=False, wraparound=False, cdivision=True
 # distutils: language = c++
-# -*- coding: utf-8 -*-
 """
 This is the variogram estimater, implemented in cython.
 """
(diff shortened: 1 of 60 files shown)


src/gstools/variogram/estimator.pyx
```diff
@@ -251,7 +250,7 @@ def unstructured(
     else:
         distance = dist_haversine
         if dim != 2:
-            raise ValueError('Haversine: dim = {0} != 2'.format(dim))
+            raise ValueError('Haversine: dim = {} != 2'.format(dim))
```
MuellerSeb, line 253: Doesn't Cython understand f-strings as well? [link](http://blog.behnel.de/posts/whats-new-in-cython-024.html)

MuellerSeb (commented):
That is what I like about OSS: people care about it! Thanks for that. One minor comment about the usage of f-strings in cython.

src/gstools/variogram/estimator.pyx
mwtoews: I've manually corrected these, since yes, Cython has good f-string support

MuellerSeb (approved):
Looks good to me! Thank you for this cleanup @mwtoews

PR #302, opened 2023-03-08
Add a Fourier generator for periodic spatial random fields

LSchueler:
This is a first attempt to include a new generator based on the Fourier method. One of its desirable properties is the periodicity of the generated fields, which can be a necessity for lower boundary conditions in atmospherical physics.

Tests are still completely missing and I think the length scales of the fields are currently not working as expected, which is probably due to the rescaling of the Fourier modes. Before commiting any more to this, I'll wait for the comments from an atmospherical physicist.

Here is a simple script to show you how to currently use the new generator:

```python
import numpy as np
import gstools as gs
import matplotlib.pyplot as pt

x = np.arange(100)
y = np.arange(100)

model = gs.Gaussian(dim=2, var=1, len_scale=30.0)
srf = gs.SRF(model, generator='Fourier', modes_truncation=[1.0, 1.0], modes_no=[80, 80], seed=3684903)

field = srf((x, y), mesh_type='structured')

srf.plot()
pt.show()
```


mschlutow:
Something's not quite right. The shape of the histogram is not Gaussian and its variance not unity. Becomes even more crooked when more modes are added.

```
import numpy as np
import gstools as gs
import matplotlib.pyplot as pt

Nx=512
Ny=256

x = np.arange(Nx)
y = np.arange(Ny)

model = gs.Gaussian(dim=2, var=1, len_scale=100.0)
srf = gs.SRF(model, generator='Fourier', modes_truncation=[1.0, 1.0], modes_no=[80, 80], seed=3684903)

field = srf((x, y), mesh_type='structured')

srf.plot()
pt.show()

pt.hist(field.flatten('C'),bins=64)
pt.show()
```
Figure 2023-03-14 114447

LSchueler:
Thanks for providing the script :+1: I guess the reason is that you only sample very few correlation lengths. I just ran your script, but with `len_scale=1.0` and this is the result:
Figure_1

mschlutow:
Thanks for the quick reply. 
I am afraid that it is not only the length scale that leads to uneven distribution.   
Figure 2023-03-14 152012
```
model = gs.Gaussian(dim=2, var=1.0, len_scale=200.0)
srf = gs.SRF(model, generator='Fourier', modes_truncation=[1.0, 1.0], modes_no=[512, 256])
```
Why is it symmetrical?

LSchueler:
Yeah, there is definitely something wrong with the length scales. Here is a script to estimate the variograms, which also gives you the actual length scale of a field. If you want to use your own field generator, simply ignore everything until line 21 and give your field to the variogram estimator as the `field` keyword in line 29.
I have to see, if I find some time this evening to check the rescaling of the Fourier modes.

```python
import numpy as np
import gstools as gs
import matplotlib.pyplot as pt

x = np.arange(0, 100, 1)
y = np.arange(0, 60, 1)

dim = 2
model = gs.Gaussian(dim=dim, var=1, len_scale=[5.0, 3.0])
# fourier method
srf = gs.SRF(
    model,
    generator='Fourier',
    modes_truncation=[1.0, 1.0],
    modes_no=[100, 60],
    seed=3684903,
)
# randomization method
# srf = gs.SRF(model, seed=3684903)

field = srf((x, y), mesh_type='structured')

x_max = 40.0
angle = 0.0

bins = np.arange(0, x_max, 2)

bin_center, dir_vario, counts = gs.vario_estimate(
    *((x, y), field, bins),
    direction=gs.rotated_main_axes(dim, angle),
    angles_tol=np.pi / 16,
    bandwidth=8,
    mesh_type="structured",
    return_counts=True,
)

model.fit_variogram(bin_center, dir_vario)
print(f'Fitted model: {model}')

fig, (ax1, ax2) = pt.subplots(1, 2, figsize=[10, 5])

ax1.scatter(bin_center, dir_vario[0], label="emp. vario in x-dir")
ax1.scatter(bin_center, dir_vario[1], label="emp. vario: in y-dir")
ax1.legend(loc="lower right")

model.plot("vario_axis", axis=0, ax=ax1, x_max=x_max, label="fit in x-dir")
model.plot("vario_axis", axis=1, ax=ax1, x_max=x_max, label="fit in y-dir")
ax1.set_title("Fitting an anisotropic model")

srf.plot(ax=ax2)
pt.show()
```

LSchueler:
Yeah, I think the final problem to solve is the scaling of the Fourier modes. This is an example, where I manually figured out the rescaling for one specific set of parameters and the (anisotropic) length scales are estimated correctly.

Figure_1

MuellerSeb:
Please rebase with main.

LSchueler:
> Please rebase with main.

Good idea :sweat_smile:

mschlutow:
`GSTools/examples/00_misc/06_fourier.py` throws error

```Reloaded modules: gstools.config, gstools.tools.export, gstools.tools.geometric, gstools.tools.special, gstools.tools, gstools.tools.misc, gstools.covmodel.plot, gstools.covmodel.tools, gstools.covmodel.fit, gstools.covmodel.base, gstools.covmodel.models, gstools.covmodel.tpl_models, gstools.covmodel, gstools.normalizer.base, gstools.normalizer.methods, gstools.normalizer.tools, gstools.normalizer, gstools.field.tools, gstools.transform.array, gstools.transform.field, gstools.transform, gstools.field.base, gstools.random.tools, gstools.random.rng, gstools.random, gstools.field.summator, gstools.field.generator, gstools.krige.tools, gstools.variogram.binning, gstools.variogram.estimator, gstools.variogram.variogram, gstools.variogram, gstools.krige.krigesum, gstools.krige.base, gstools.krige.methods, gstools.krige, gstools.field.cond_srf, gstools.field.upscaling, gstools.field.srf, gstools.field, gstools._version, gstools, gstools.field.plot
/home/mschlutow/GSTools/src/gstools/field/plot.py:407: UserWarning: Matplotlib is currently using module://matplotlib_inline.backend_inline, which is a non-GUI backend, so cannot show the figure.
  fig.show()
Traceback (most recent call last):

  File ~/anaconda3/envs/nansen/lib/python3.9/site-packages/spyder_kernels/py3compat.py:356 in compat_exec
    exec(code, globals, locals)

  File ~/GSTools/examples/00_misc/06_fourier.py:55
    modes = [np.arange(0, modes_cutoff[d], modes_delta[d]) for d in 2]

TypeError: 'int' object is not iterable```

**Comment by LSchueler** [comment]
Sorry, it was late last evening... I just fixed that problem.

**Comment by LSchueler** [comment]
Wow, this was a looong journey. But I think we are finally there. We did a lot of testing in the background and the Fourier method seems to be working robustly now.
@MuellerSeb Do you have time for a review?

**Comment by MuellerSeb** [comment]
Thanks @LSchueler for the great work. Will have a look as soon as I find the time!
Could you update the branch with main?

File: src/gstools/field/generator.py [hunk 1]
```diff
@@ -1,4 +1,4 @@
-"""
+"""gene
```
MuellerSeb, line 1: Remove.

src/gstools/field/generator.py
```diff
@@ -271,7 +279,7 @@ def update(self, model=None, seed=np.nan):
                 raise ValueError(
                     "gstools.field.generator.RandMeth: no 'model' given"
                 )
-        # if the user tries to trick us, we beat him!
+        # if the user tries to trick us, we beat them!
```
MuellerSeb, line 282: Good catch.

src/gstools/field/generator.py
```diff
@@ -532,3 +540,349 @@ def _create_unit_vector(self, broadcast_shape, axis=0):
...
+        * :math:`k_i` : the equidistant Fourier grid
+    """
+
+    def __init__(
+        self,
+        model,
+        mode_no,
+        period,
+        seed=None,
+        verbose=False,
```
MuellerSeb, line 592: verbose is gone now I think.

src/gstools/field/generator.py
```diff
@@ -532,3 +540,349 @@ def _create_unit_vector(self, broadcast_shape, axis=0):
...
+        self.update(period=period)
+
+    @property
+    def verbose(self):
+        """:class:`bool`: Verbosity of the generator."""
+        return self._verbose
+
+    @verbose.setter
+    def verbose(self, verbose):
+        self._verbose = bool(verbose)
```
MuellerSeb, line 879: verbose is gone now.

src/gstools/field/generator.py
```diff
@@ -532,3 +540,349 @@ def _create_unit_vector(self, broadcast_shape, axis=0):
...
+        # if the user tries to trick us, we beat them!
+        elif model is None and np.isnan(seed):
+            if (
+                isinstance(self._model, CovModel)
+                and self._z_1 is not None
+                and self._z_2 is not None
+                and self._spectrum_factor is not None
+            ):
+                if self.verbose:
+                    print("Fourier.update: Nothing will be done...")
```
MuellerSeb, line 738: verbose is gone now.

src/gstools/field/srf.py
```diff
@@ -142,6 +149,11 @@ def __call__(
         field : :class:`numpy.ndarray`
             the SRF
         """
+        if isinstance(self.generator, Fourier) and mesh_type != "structured":
+            raise ValueError(
+                "SRF: Fourier generator only defined for "
+                'mesh_type == "structured".'
+            )
```
MuellerSeb, line 156: Does it really need to be structured? I think it is passed unstructured to the generator either way and since the period is not determined from the pos tuple anymore, this check could be dropped.

MuellerSeb (commented):
I didn't test it yet and just had a first look. There are some minor things already but it looks good overall! Great work.

src/gstools/field/srf.py
LSchueler: For the Fourier method to work, the spatial grid has to be regular, so this is a convenient way to ensure that. Otherwise, I could imagine that users might run into quite obscure problems with the field generation. Maybe we could add that feature, when it is really needed?

LSchueler:
I merged the main branch and I also implemented the Fourier method in GSTools-Core, so included that import in GSTools, if GSTools-Core is available.

src/gstools/field/srf.py
MuellerSeb: But the generator gets `iso_pos` which are always unstructured and from going through the summator I don't see where it matters that the original pos tuple was defining a structured grid. Isn't the whole point about the Fourier generator, that it is periodic? And since the periods in each main direction are now given independently from the field, I think this should just work.

Imagine the follwing scenario: We have a 2x2 grid. If we only pass two points from that (top-left, bottom-right) as an unstructured grid to the summator I should get the same results for these two points like within the structured 2x2 grid. Since the summator calulates the results for each point independently. Scaling this example, for an arbitrary unstructured grid you could generate a structured grid containing all points from the unstructured one. Now generate the field and cut out the unstructured grid. This should again be the same field as passing the original unstructured field. So I still think it is fine to use unstructured grids.

Or are the other restrictions? Like that the fields needs to be wihtin the first period or something similar?

.github/workflows/main.yml
```diff
@@ -124,7 +124,7 @@ jobs:
         run: |
           pip install "numpy${{ matrix.ver.np }}" "scipy${{ matrix.ver.sp }}"
           python -m pytest --cov gstools --cov-report term-missing -v tests/
-          python -m coveralls --service=github
+          #python -m coveralls --service=github
```
MuellerSeb, line 127: Why is this commented out?

.github/workflows/main.yml
```diff
@@ -48,9 +48,9 @@ jobs:
         run: |
           python -m pylint src/gstools/
 
-      - name: cython-lint check
-        run: |
-          cython-lint src/gstools/
+      #- name: cython-lint check
+        #run: |
+          #cython-lint src/gstools/
```
MuellerSeb, line 53: This should be solved with #343

src/gstools/field/srf.py
LSchueler: You are completely right that the Fourier method works with a regular (Euclidean) mesh passed to GSTools as what we call an unstructured mesh. However, the Fourier method is only defined on regular meshes. By throwing that exception, we can easily prevent problems, which will be hard to catch later, as they are pretty subtle. Of course, with a structured mesh, you can still create meshes, with non-equidistant steps, but I guess that happens very seldomly.
Regarding your last sentence: No, the periodicity can be larger and smaller than the domain size.
But in general, the periodicity breaks on unstructured meshes, which makes the keyword `period` a bit strange. Here is the output of a slightly adapted version of your tutorial with the Fourier method. The periodicity should be about half the domain size.

unstruct_fourier

I don't know, I could drop the exception and print a warning, about the features of the unstructured mesh or put a very clear warning in the docs?

.github/workflows/main.yml
LSchueler: The coveralls service was down once and stopped the runners completely, but I wanted to check if they complete successfully.
But yeah, that change shouldn't make it into main.

src/gstools/field/srf.py
LSchueler: By the way, at some point we should find an alternative to meshzoo ;-)

LSchueler: Okay, after some testing, the Fourier method *does* indeed work on non-equidistant meshes, so dropping that exception.

.github/workflows/main.yml
LSchueler: I'll solve that in a different PR.

LSchueler: That's going to be solved in PR #354.

src/gstools/field/generator.py
```diff
@@ -532,3 +538,332 @@ def _create_unit_vector(self, broadcast_shape, axis=0):
...
+
+        * :math:`S` : spectrum of the covariance model
+        * :math:`Z_{j,i}` : random samples from a normal distribution
+        * :math:`k_i` : the equidistant Fourier grid
+    """
+
+    def __init__(
+        self,
+        model,
+        mode_no,
```
MuellerSeb, line 584: Could we provide a default value for `mode_no` like with the RandMeth class?
Maybe something like `model.dim * [32]` following your examples?

src/gstools/field/generator.py
```diff
@@ -532,3 +538,332 @@ def _create_unit_vector(self, broadcast_shape, axis=0):
...
+        if period is not None:
+            if len(period) != tmp_model.dim:
+                raise ValueError(
+                    "Fourier: Dimension mismatch in argument period."
+                )
+            self._period = np.array(period)
+            self._delta_k = 2.0 * np.pi / self._period
+            if mode_no is None:
+                self._set_modes(self._mode_no, tmp_model)
+        if mode_no is not None:
```
MuellerSeb, line 688: There should be an else branch checking if `self._mode_no` is None to ensure mode_no is given.

src/gstools/field/generator.py
```diff
@@ -532,3 +538,332 @@ def _create_unit_vector(self, broadcast_shape, axis=0):
...
+            # just update the seed, if its a new one
+            elif seed is None or not np.isnan(seed):
+                self.seed = seed
+        # or just update the seed, when no model is given
+        elif model is None and (seed is None or not np.isnan(seed)):
+            if isinstance(self._model, CovModel):
+                self.seed = seed
+            else:
+                raise ValueError(
+                    "gstools.field.generator.RandMeth: no 'model' given"
```
MuellerSeb, line 714: Copypasta: Should be `Fourier` instead of `RandMeth`

src/gstools/field/generator.py
MuellerSeb: And I would like some syntactic sugar, to be able to only give one number that is then used for all dimensions depending on the cov-model.

MuellerSeb: Then we could use `mode_no=32` as default ;-)

src/gstools/field/generator.py
```diff
@@ -532,3 +538,332 @@ def _create_unit_vector(self, broadcast_shape, axis=0):
...
+        * :math:`S` : spectrum of the covariance model
+        * :math:`Z_{j,i}` : random samples from a normal distribution
+        * :math:`k_i` : the equidistant Fourier grid
+    """
+
+    def __init__(
+        self,
+        model,
+        mode_no,
+        period,
```
MuellerSeb, line 585: Same as for mode_no, the possibility to give a single number that is distributed to all dimensions would be cool!

MuellerSeb: Also I would put `period` before `mode_no` in the signature since it's the more important argument.

MuellerSeb (commented):
I tested it and think this is fine! Some comments about the Generator.

One thing I noticed is that variance is too low when estimated but I think this is fine, since ergodicity is hard/impossible to achieve with periodic fields and that it is in the nature of things that field variance is lower than the defined variance of the used model.

My observation is: the lower the period, the lower the variance. And this makes perfectly sense.

src/gstools/field/generator.py
LSchueler: good idea, implemented.

src/gstools/field/generator.py
LSchueler: I'm not sure, if I understand you correctly, but it should be a valid way to call the `update` , e.g. if you only want to change the `period`, but not the `mode_no`. That case is already handled by the code.

src/gstools/field/generator.py
LSchueler: Thanks for pointing that out.

src/gstools/field/generator.py
LSchueler: Thanks for the good idea, implemented.

.github/workflows/main.yml
LSchueler: PR #354 should be merged first and then I can update this branch with the fixes.

src/gstools/field/generator.py
```diff
@@ -532,3 +538,326 @@ def _create_unit_vector(self, broadcast_shape, axis=0):
...
+                raise ValueError("Fourier: Value has to be provided")
+            r = default_value
+        r = np.array(r, dtype=dtype)
+        r = np.atleast_1d(r)[:dim]
+        if len(r) > dim:
+            raise ValueError(f"Fourier: len(values) <= {dim=} not fulfilled")
+        # fill up values with values[-1], such that len()==dim
+        if len(r) < dim:
+            r = np.pad(r, (0, dim - len(r)), "edge")
+        return r
```
MuellerSeb, line 780: This could be a tool function not a method.

examples/01_random_field/08_fourier.py
```diff
@@ -0,0 +1,38 @@
...
+# scale which is roughly half the size of the grid.
+model = gs.Gaussian(dim=2, var=1, len_scale=200)
+
+# Next, we hand the cov. model to the spatial random field class
+# and set the generator to `Fourier`. The `mode_no` argument sets the number of
+# Fourier modes per dimension. The argument `period` is set to the domain size.
+srf = gs.SRF(
+    model,
+    generator="Fourier",
+    mode_no=[32, 32],
```
MuellerSeb, line 29: For the simple example I would use `mode_no=32`

examples/01_random_field/08_fourier.py
```diff
@@ -0,0 +1,38 @@
...
+random fields. The Fourier method comes naturally with the property of
+periodicity, so we'll use it to create the random field.
+"""
+
+import numpy as np
+
+import gstools as gs
+
+# We start off by defining the spatial grid.
+L = np.array((500, 500))
```
MuellerSeb, line 15: Since L has the same values in each direction we could simplify this example by using a `size=500` variable and also pass this as a single period argument.

examples/01_random_field/09_fourier_trans.py
```diff
@@ -0,0 +1,45 @@
...
+Building on the precious example, we are now going to generate periodic
+spatial random fields with a transformation applied, resulting in a level set.
+"""
+
+import numpy as np
+
+import gstools as gs
+
+# We start off by defining the spatial grid.
+L = np.array((500, 500))
```
MuellerSeb, line 14: Maybe use different periods for different directions in this example?

examples/01_random_field/09_fourier_trans.py
```diff
@@ -0,0 +1,45 @@
...
+x = np.linspace(0, L[0], 300)
+y = np.linspace(0, L[1], 200)
```
MuellerSeb, line 16: Maybe use double the period to see the periodicity in the generated field? (`2*L[0]` and `2*L[1]`)

MuellerSeb (commented):
Like it! some minor comments.

src/gstools/field/generator.py
```diff
@@ -532,3 +538,326 @@ def _create_unit_vector(self, broadcast_shape, axis=0):
...
+       \sum_{i=1}^{N}\sqrt{2S(k_{i})\Delta k}\left(
+       Z_{1,i}\cdot\cos\left(\left\langle k_{i},x\right\rangle \right)+
+       Z_{2,i}\cdot\sin\left(\left\langle k_{i},x\right\rangle \right)
+       \right)
+
+    where:
+
+        * :math:`S` : spectrum of the covariance model
+        * :math:`Z_{j,i}` : random samples from a normal distribution
+        * :math:`k_i` : the equidistant Fourier grid
```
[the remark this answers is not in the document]
LSchueler: Yeah, that's still true, those are the wave numbers and they have to be equidistant.

examples/01_random_field/08_fourier.py
LSchueler: Good idea, implemented.

examples/01_random_field/08_fourier.py
LSchueler: Good idea, implemented.

examples/01_random_field/09_fourier_trans.py
LSchueler: Good idea, implemented.

examples/01_random_field/09_fourier_trans.py
LSchueler: I've used half the domain length in y-direction.

src/gstools/field/generator.py
LSchueler: Now that we have the `Fourier.__init__` argument `mode_no` set to a default value, this shouldn't be possible anymore, right? - I mean, except someone tries really hard ;-)

> or on the verge to start using static typing in python ;-)

No way! :rofl: You just proposed to make it possible to do `period=200` *and* `period=[200, 200]`.

MuellerSeb: OK, you are right. Btw, that is totally fine with typing and would be `Union[int, List[int]]` ;-)

MuellerSeb (approved):
LGTM

LSchueler:
I can't believe it, after nearly 1 1/2 years, this is ready! Thanks for the reviews.

Issue #282, opened 2022-11-07
Feature Enhancement: Unit selection and max_distance with Variogram where latlon = True


Megil-Zesty:
When defining a variogram with latlon=True, I would love to be able to set max_distance directly in  gs.vario_estimate() as well as the units for the distance. Then, when plotting the variogram data, I could see semivariance vs distance.

MuellerSeb:
Could you provide a code-example? I don't think I fully understand your problem.

MuellerSeb:
#308 solves at least the units issue for the CovModel representation, that can now be defined with the new `geo_scale` attribute of the CovModel. You can choose between `KM_SCALE` (earth radius in km as reference), `DEGREE_SCALE` (length scale in degrees) and `RADIAN_SCALE` (default, length scale in radians). To use other distance units, you could multiply conversion factors to `KM_SCALE`, like `1000 * KM_SCALE` to have a meter scale.

The `vario_estimate` routine will always use radians for the great circle distance. To have a better plotting, you could multiply the bin centers with one of the scales given above to have another unit for the x-axis.

For the other problem I assume you mean the remaining arguments of the [standard_bins](https://geostat-framework.readthedocs.io/projects/gstools/en/v1.4.1/generated/generated/gstools.variogram.standard_bins.html) routine, namely `bin_no` and `max_dist`.
This seems like a useful extension. I will take a look.

MuellerSeb:
`geo_scale` was added to `vario_estimate` to select the units of the bins with #308

MuellerSeb:
Also additional keywords will be forwarded to `standard_bins`, so this will be solved.

Issue #140, opened 2021-03-13
xarray interface


MuellerSeb:
We should add interfaces to xarray wihtin the Field-class:

- `xarray` method: generate field on xarray
- `to_xarray` export field to a `xarray`

We would need to think about call signature for `xarray` (selecting dims, name, determining struct/unstruct)

PR #308, opened 2023-06-06
Add 'temporal', 'spatial_dim' and 'geo_scale' attributes to Covmodel

### LINKED ISSUE #223: [Suggestion] Support handling geographical data with a time component
Hi there, I recently came across the GSTools package and think that this could be perfect in helping me accomplish my goal -- generating random fields according to the correlation lengths of spatio-temporal data (e.g. precipitation data with lat, lon, time as the dimensions). I especially appreciate the option `latlon=True` since it is not trivial to find a correlation length in degree when 1 degree can mean very different distances at different latitudes. I found that my workflow is still rather complex due to the limitation that the input data has to be 2D in order to use the latlon option (detailed description of the workflow below). Since processing spatio-temporal data is a fairly common problem, I was wondering if you have considered adding support for using 3D or 4D data while two of the dimensions are lat/lon? Alternatively, if there is a working pattern for handling spatio-temporal data with the current functionality, an example would be greatly appreciated. 

Currently, my workflow consists of two steps: 
1) Find the correlation lengths of the source data. To do this I'm first getting the spatial correlation length using `latlon=True` option (with each time point being different examples), then find the temporal correlation length separately (with each location being different examples); 
2) Generate the random fields using this correlation length. To do this, I first project the lat/lon grid I want into an equidistant projection and also scale the spatial correlation length into meters. I then generate a 3D random field with the correlation lengths in units of (meters, meters, day). Finally, I reproject the 3D random field back into lat/lon grid. The reprojections are needed since the latlon option is not available if I'd like to generate random fields that are more than 2D. 

Thanks for working on this great package and looking forward to hearing your thoughts!

MuellerSeb:
Closes #223, #282

This PR adds a boolean `temporal` and `spatial_dim` attribute to the CovModel class as well as `geo_scale` to CovModel and vario_estimate.

Setting `temporal=True` indicates that the model should be seen as a metric spatio-temporal model. This was possible before by increasing the model dimension by one but this feature basically enables using spatio-temporal model with latlon coordinates. `spatial_dim` was added so the spatial dimension could be set explicitly and not implicitly by increasing only `dim` by 1.

Also a `geo_scale` attribute was added to be able to get meaningful length-scales and bins for latlon models. Before we always used the `rescale` attribute to do that, but that was rather a hack. Now both can be used to set the sphere radius and to rescale the length-scale. In addition to that I added `KM_SCALE`, `DEGREE_SCALE` and `RADIAN_SCALE` constant that can be used as radius to get a length-scale in km, degree or radian as some people expect the length-scale to be in degree when working latlon coordinates.

Another dimension indicator attribute was added to the CovModel class:
- `spatial_dim`: The truly spatial dimension of the model (2 for latlon, otherwise `dim` or `dim-1` (if `temporal=True`))
  - in contrast to `dim`: the internal model dimension (4 for `latlon=True` and `temporal=True` for example)
  - in contrast to `field_dim`: this is the parametric dimension of the generated field (3 for `latlon=True` and `temporal=True`, 2 for `latlon=True` and `temporal=False` for example)

Works like this:
```python
import numpy as np
import gstools as gs

model = gs.Matern(
    temporal=True,
    latlon=True,
    len_scale=[1000, 100],
    geo_scale=gs.KM_SCALE,
)

lat = lon = np.linspace(-80, 81, 50)
time = np.linspace(0, 777, 50)
srf = gs.SRF(model, seed=1234)
field = srf.structured((lat, lon, time))
srf.plot()
```
Figure_1
Figure_2

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,7 +24,7 @@ All notable changes to **GSTools** will be documented in this file.
 - `vario_estimate` now forwards additional kwargs to `standard_bins` (`bin_no`, `max_dist`) [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
 
 ### Changes
-- CovModels expect all arguments by keyword now (except `dim`) [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
+- CovModels expect special arguments by keyword now [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
 - always use f-strings internally [#283](https://github.com/GeoStat-Framework/GSTools/pull/283)
 
 ### Bugfixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,34 @@
 
 All notable changes to **GSTools** will be documented in this file.
 
+## [Unreleased] - ? - 2023-?
+
+### Enhancements
+- added `temporal` flag to CovModel to explicitly specify spatio-temporal models [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
+  - rotation between spatial and temporal dimension will be ignored
+  - added `spatial_dim` to CovModel to explicitly set spatial dimension for spatio-temporal models
+    - if not using `spatial_dim`, the provided `dim` needs to include the possible temporal dimension
+    - `spatial_dim` is always one less than `field_dim` for spatio-temporal models
+  - also works with `latlon=True` to have a spatio-temporal model with geographic coordinates
+  - all plotting routines respect this
+  - the `Field` class now has a `temporal` attribute which forwards the model attribute
+  - automatic variogram fitting in kriging classes for `temporal=True` and `latlon=True` will raise an error
+- added `geo_scale` to CovModel to have a more consistent way to set the units of the model length scale for geographic coordinates [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
+  - no need to use `rescale` for this anymore (was rather a hack)
+  - added `gs.KM_SCALE` which is the same as `gs.EARTH_RADIUS` for kilometer scaling
+  - added `gs.DEGREE_SCALE` for great circle distance in degrees
+  - added `gs.RADIAN_SCALE` for great circle distance in radians (default and previous behavior)
+  - yadrenko variogram respects this and assumes the great circle distances is given in the respective unit
+  - `vario_estimate` also has `geo_scale` now to control the units of the bins
+- `vario_estimate` now forwards additional kwargs to `standard_bins` (`bin_no`, `max_dist`) [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
+
+### Changes
+- CovModels expect all arguments by keyword now (except `dim`) [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
+- always use f-strings internally [#283](https://github.com/GeoStat-Framework/GSTools/pull/283)
+
+### Bugfixes
+- latex equations were not rendered correctly in docs [#290](https://github.com/GeoStat-Framework/GSTools/pull/290)
+
 
 ## [1.4.1] - Sassy Sapphire - 2022-11
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,16 +5,16 @@ All notable changes to **GSTools** will be documented in this file.
 ## [Unreleased] - ? - 2023-?
 
 ### Enhancements
-- added `temporal` flag to CovModel to explicitly specify spatio-temporal models [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
+- added `temporal` flag to `CovModel` to explicitly specify spatio-temporal models [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
   - rotation between spatial and temporal dimension will be ignored
-  - added `spatial_dim` to CovModel to explicitly set spatial dimension for spatio-temporal models
+  - added `spatial_dim` to `CovModel` to explicitly set spatial dimension for spatio-temporal models
     - if not using `spatial_dim`, the provided `dim` needs to include the possible temporal dimension
     - `spatial_dim` is always one less than `field_dim` for spatio-temporal models
   - also works with `latlon=True` to have a spatio-temporal model with geographic coordinates
   - all plotting routines respect this
   - the `Field` class now has a `temporal` attribute which forwards the model attribute
   - automatic variogram fitting in kriging classes for `temporal=True` and `latlon=True` will raise an error
-- added `geo_scale` to CovModel to have a more consistent way to set the units of the model length scale for geographic coordinates [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
+- added `geo_scale` to `CovModel` to have a more consistent way to set the units of the model length scale for geographic coordinates [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
   - no need to use `rescale` for this anymore (was rather a hack)
   - added `gs.KM_SCALE` which is the same as `gs.EARTH_RADIUS` for kilometer scaling
   - added `gs.DEGREE_SCALE` for great circle distance in degrees
@@ -53,7 +53,7 @@ All notable changes to **GSTools** will be documented in this file.
 - better support for custom generators [#250](https://github.com/GeoStat-Framework/GSTools/pull/250) [#259](https://github.com/GeoStat-Framework/GSTools/pull/259)
 - add `valid_value_types` class variable to all field classes [#250](https://github.com/GeoStat-Framework/GSTools/pull/250)
 - PyKrige: fix passed variogram in case of latlon models [#254](https://github.com/GeoStat-Framework/GSTools/pull/254)
-- add bounds checks for optional arguments of CovModel when resetting by class attribute [#255](https://github.com/GeoStat-Framework/GSTools/pull/255)
+- add bounds checks for optional arguments of `CovModel` when resetting by class attribute [#255](https://github.com/GeoStat-Framework/GSTools/pull/255)
 - minor coverage improvements [#255](https://github.com/GeoStat-Framework/GSTools/pull/255)
 - documentation: readability improvements [#257](https://github.com/GeoStat-Framework/GSTools/pull/257)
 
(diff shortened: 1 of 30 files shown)


src/gstools/covmodel/base.py
```diff
@@ -530,14 +542,14 @@ def isometrize(self, pos):
         """Make a position tuple ready for isotropic operations."""
         pos = np.asarray(pos, dtype=np.double).reshape((self.field_dim, -1))
         if self.latlon:
-            return latlon2pos(pos)
+            return latlon2pos(pos, time=self.time)
```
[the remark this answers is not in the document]
MuellerSeb: This behaves strange. I think we should add a `radius` attribute to CovModel to prevent the `rescale` hack to get a meaningful len_scale. At the moment the len_scale is given on the unit-sphere and that doesn't really work with the time-dimension and the associated len_scale.

MuellerSeb:
@LSchueler Maybe `radius` is a bit misleading, since it is a rather technical term with respect to the internal representation of the yadrenko-variogram. May we could call this attribute `geo_scale` with is a bit more self-descriptive. What do you think?

MuellerSeb:
@LSchueler I like the `geo_scale` attribute. Next question, following #282 is, if we want to add `geo_scale` also to the `vario_estimate` routine so the bin_centers have a meaningful unit?

MuellerSeb:
vario_estimate was update with `geo_scale` and `**std_bins` arguments. Now everything feels really straight forward.

MuellerSeb:
@LSchueler we may disable rotation between spatial dimensions and time if `time=True`. Otherwise I think the metric spatio-temporal model is violated.

LSchueler:
Sorry for responding so late. I was sick.

This looks interesting!
Just a few very brief comments:
- I like the keyword `geo_scale`
- however, I think `time` is a bit too generic. Although it's a bit longer, I think `spatio-temporal` makes it much clearer what the keyword is for.
- I fear that maintenance would become even more difficult if we follow the CF-conventions for spatio-temporal fields. What about providing a convenience function or wrapper returning and/or receiving the data following CF-convention.
- Disabling the rotation is a good idea until we figure out the exact meaning of rotation of spatio-temporal fields.

MuellerSeb:
Or only `temporal`?

examples/09_spatio_temporal/03_geographic_coordinates.py
```diff
@@ -0,0 +1,36 @@
...
+First we setup a model, with ``latlon=True`` and ``time=True``,
+to get the associated spatio-temporal Yadrenko model.
+
+In addition, we will use the earth radius provided by :any:`EARTH_RADIUS`
+as ``geo_scale`` to have a meaningful length scale in km.
+
+To generate the field, we simply pass ``(lat, lon, time)`` as the position tuple
+to the :any:`SRF` class.
+
+The anisotropy factor of `0.1` (days/km) will result in a time length-scale of `100` days.
```
LSchueler, line 17: I think I would find it more intuitive, if a list of length scales is provided, instead of an anisotropy-factor.

src/gstools/covmodel/base.py
```diff
@@ -100,9 +102,21 @@ class CovModel:
...
         As a consequence, `dim` will be set to `3` and anisotropy will be
-        disabled. `rescale` can be set to e.g. earth's radius,
+        disabled. `geo_scale` can be set to e.g. earth's radius,
         to have a meaningful `len_scale` parameter.
         Default: False
+    geo_scale : :class:`float`, optional
+        Geographic scaling in case of latlon coordinates to get a meaningful length
+        scale. By default, len_scale is assumed to be in radians with latlon=True.
+        Can be set to :any:`EARTH_RADIUS` to have len_scale in km or
+        :any:`DEGREE_SCALE` to have len_scale in degree.
```
LSchueler, line 112: I find this confusing. By just looking at it, I would assume that the `len_scale` is multiples of `geo_scale` or something like that (as `EARTH_RADIUS` is only one possible radius). Maybe reformulate that the physical unit is km as soon as a radius is set.

src/gstools/variogram/variogram.py
```diff
@@ -222,6 +225,15 @@ def vario_estimate(
     fit_normalizer : :class:`bool`, optional
         Whether to fit the data-normalizer to the given (detrended) field.
         Default: False
+    geo_scale : :class:`float`, optional
+        Geographic scaling in case of latlon coordinates to get meaningful bins.
+        By default, bins are assumed to be given in radians with latlon=True.
+        Can be set to :any:`EARTH_RADIUS` to have units in km or
+        :any:`DEGREE_SCALE` to have units in degree.
+        Default: :any:`RADIAN_SCALE`
```
LSchueler, line 233: See above comment

src/gstools/covmodel/base.py
LSchueler: I actually meant the formulation of the documentation, but also only using `KM_SCALE` is a good idea :+1:

examples/09_spatio_temporal/03_geographic_coordinates.py
MuellerSeb: Nevermind, it works... but why?!

MuellerSeb: Wow, just a solid logic for len_scale and anis implemented in the past. Will use `len_scale=[1000, 100]` in this example.

src/gstools/variogram/variogram.py
MuellerSeb: Done

src/gstools/covmodel/base.py
MuellerSeb: Done

examples/09_spatio_temporal/03_geographic_coordinates.py
MuellerSeb: Done

MuellerSeb:
@LSchueler all fixes implemented. I would also vote against the CF conventions and maybe have that with #140

MuellerSeb:
@LSchueler Only question I still have is, if `dim` should really be automatically increased by 1 if `temporal=True`.

For example if you want a model with 3 spatial dimensions and a time dimension, which option looks better to you:

```python
model = gs.Gaussian(dim=3, temporal=True)
```
or
```python
model = gs.Gaussian(dim=4, temporal=True)
```
In both cases `model.dim` would be 4 and `model.spatial_dim` would be 3.

MuellerSeb:
I added `spatial_dim` to solve this. `dim` now always really refers to the model dimension. `spatial_dim` could be used instead, when defining a spatio-temporal model:
```python
model = gs.Gaussian(temporal=True, spatial_dim=3)
model.dim == 4
model.spatial_dim == 3
```
This is equal to:
```
model = gs.Gaussian(temporal=True, dim=4)
```

examples/08_geo_coordinates/00_field_generation.py
```diff
@@ -8,15 +8,19 @@
 First we setup a model, with ``latlon=True``, to get the associated
 Yadrenko model.
 
-In addition, we will use the earth radius provided by :any:`EARTH_RADIUS`,
-to have a meaningful length scale in km.
+In addition, we will use a kilometer scale provided by :any:`KM_SCALE`
+as ``geo_scale`` to have a meaningful length scale in km.
+By default the length scale would be given in radians (:any:`RADIAN_SCALE`).
+A third option is a length scale in degrees (:any:`DEGREE_SCALE`).
```
LSchueler, line 14: :+1:

examples/09_spatio_temporal/03_geographic_coordinates.py
```diff
@@ -0,0 +1,37 @@
...
+    var=1,
+    len_scale=[1000, 100],
+    geo_scale=gs.KM_SCALE,
+)
+
+lat = lon = np.linspace(-80, 81, 50)
+time = np.linspace(0, 777, 50)
+srf = gs.SRF(model, seed=1234)
+field = srf.structured((lat, lon, time))
+srf.plot()
```
LSchueler, line 37: Nice example!

src/gstools/variogram/variogram.py
```diff
@@ -250,18 +263,18 @@ def vario_estimate(
...
     if masked and np.all(mask):
-        bin_centres = np.empty(0) if bin_edges is None else bin_centres
-        estimates = np.zeros_like(bin_centres)
+        bin_center = np.empty(0) if bin_edges is None else bin_center
+        estimates = np.zeros_like(bin_center)
         if return_counts:
-            return bin_centres, estimates, np.zeros_like(estimates, dtype=int)
-        return bin_centres, estimates
+            return bin_center, estimates, np.zeros_like(estimates, dtype=int)
+        return bin_center, estimates
```
LSchueler, line 277: I find it quite confusing that it's the plural of `bin_edges`, but now the singular of `bin_center`.

src/gstools/variogram/variogram.py
```diff
@@ -371,7 +389,7 @@ def vario_estimate(
         if dir_no == 1:
             estimates, counts = estimates[0], counts[0]
     est_out = (estimates, counts)
-    return (bin_centres,) + est_out[: 2 if return_counts else 1] + norm_out
+    return (bin_center,) + est_out[: 2 if return_counts else 1] + norm_out
```
LSchueler, line 392: See above

LSchueler (changes requested):
Nice work! I think you are nearly there.

CHANGELOG.md
```diff
@@ -2,6 +2,34 @@
...
+  - no need to use `rescale` for this anymore (was rather a hack)
+  - added `gs.KM_SCALE` which is the same as `gs.EARTH_RADIUS` for kilometer scaling
+  - added `gs.DEGREE_SCALE` for great circle distance in degrees
+  - added `gs.RADIAN_SCALE` for great circle distance in radians (default and previous behavior)
+  - yadrenko variogram respects this and assumes the great circle distances is given in the respective unit
+  - `vario_estimate` also has `geo_scale` now to control the units of the bins
+- `vario_estimate` now forwards additional kwargs to `standard_bins` (`bin_no`, `max_dist`) [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
+
+### Changes
+- CovModels expect special arguments by keyword now [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
```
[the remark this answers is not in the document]
MuellerSeb: Done

src/gstools/variogram/variogram.py
MuellerSeb: renamed

src/gstools/variogram/variogram.py
MuellerSeb: renamed

CHANGELOG.md
```diff
@@ -2,6 +2,34 @@
 
 All notable changes to **GSTools** will be documented in this file.
 
+## [Unreleased] - ? - 2023-?
+
+### Enhancements
+- added `temporal` flag to CovModel to explicitly specify spatio-temporal models [#308](https://github.com/GeoStat-Framework/GSTools/pull/308)
+  - rotation between spatial and temporal dimension will be ignored
+  - added `spatial_dim` to CovModel to explicitly set spatial dimension for spatio-temporal models
```
LSchueler, line 10: Two more naked CovModels ;-)

MuellerSeb: fixed

LSchueler (approved):
:-D Wow, my review wasn't very thorough, was it?

LGTM!

PR #318, opened 2023-06-22
tests: do setup and teardown only once per test-class

MuellerSeb:
This should speed-up the tests by only doing the setups/teardowns once per class.


MuellerSeb:
OK, seems to be a dumb idea, since some tests are altering the setups.

PR #323, opened 2023-07-15
Use integer literals where appropriate

eumiro:
Python 2 needed explicit float literals in order to do the correct math. In Python 3 this is not needed anymore. Using integers describes better the intention of the code.

This PR only changes floats that end with `.0` to integers where appropriate.


LSchueler:
Hi eumiro,

thank you for your PR.
Sometimes it makes code easier to read if float literals are explicitly used. Therefore, I don't think that converting every integer valued float to an integer literal is the right solution for GSTools and I'll close this PR.

PR #332, opened 2023-11-24
added estimator madogram

ibkvictor:
added estimator F - madogram

```
|f_diff|
```

### DIFF
--- a/src/gstools/variogram/estimator.pyx
+++ b/src/gstools/variogram/estimator.pyx
@@ -96,6 +96,9 @@ cdef inline double estimator_matheron(const double f_diff) nogil:
 cdef inline double estimator_cressie(const double f_diff) nogil:
     return sqrt(fabs(f_diff))
 
+cdef inline double estimator_madogram(const double f_diff) nogil:
+    return fabs(f_diff)
+
 ctypedef double (*_estimator_func)(const double) nogil
 
 cdef inline void normalization_matheron(
@@ -151,6 +154,8 @@ cdef _estimator_func choose_estimator_func(str estimator_type):
     cdef _estimator_func estimator_func
     if estimator_type == 'm':
         estimator_func = estimator_matheron
+    elif estimator_type == 'f':
+        estimator_func = estimator_madogram
     else:  # estimator_type == 'c'
         estimator_func = estimator_cressie
     return estimator_func
--- a/src/gstools/variogram/variogram.py
+++ b/src/gstools/variogram/variogram.py
@@ -53,6 +53,8 @@ def _set_estimator(estimator):
     """Translate the verbose Python estimator identifier to single char."""
     if estimator.lower() == "matheron":
         cython_estimator = "m"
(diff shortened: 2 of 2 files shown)


src/gstools/variogram/variogram.py
```diff
@@ -157,6 +159,7 @@ def vario_estimate(
 
             * "matheron": the standard method of moments of Matheron
             * "cressie": an estimator more robust to outliers
+            * "madogram": an esitmator for Madogram
```
LSchueler, line 162: Typo

src/gstools/variogram/estimator.pyx
```diff
@@ -151,6 +154,8 @@ cdef _estimator_func choose_estimator_func(str estimator_type):
     cdef _estimator_func estimator_func
     if estimator_type == 'm':
         estimator_func = estimator_matheron
+    elif estimator_type == 'f':
```
LSchueler, line 157: Maybe the char 'a' for absolute might be more intuitive or what is the reason you chose 'f'?

src/gstools/variogram/variogram.py
```diff
@@ -53,6 +53,8 @@ def _set_estimator(estimator):
     """Translate the verbose Python estimator identifier to single char."""
     if estimator.lower() == "matheron":
         cython_estimator = "m"
+    elif estimator.lower() == "madogram":
+        cython_estimator = "f"
```
LSchueler, line 57: See comment above.

LSchueler (changes requested):
Hi Victor,

I am truly sorry for keeping you waiting so long for a response! I like the idea of supporting the madogram. But before we can merge your PR, there are a few small things I would like to see changed.

The new functionality is not mentioned anywhere in the docstrings. Could you add some text to the docstring of the `vario_estimate` function in the file variogram.py? And could you also add a few simple unit tests?
Apart from those two points and the very small comments I made inline, I think we are nearly good to go. Thanks again!

Cheers,
Lennart

src/gstools/variogram/variogram.py
ibkvictor: resolved

src/gstools/variogram/estimator.pyx
ibkvictor: It from the name `f-madogram`, since the `matheron` estimator already took, `m`.

Here is a paper using the estimator.
paper

LSchueler:
Thank you for the changes.
However, now there is an inconsistency between the argument to choose your estimator, which is "madogram" and the docstring describing the estimator, which is "f-madogram". Please fix that.
And another problem in the docstring is the missing references. Have a look at the last lines of the docstring

```
    References
    ----------
    .. [Webster2007] Webster, R. and Oliver, M. A.
           "Geostatistics for environmental scientists.",
           John Wiley & Sons. (2007)
```

You should include your references there too and then you can link them via `[Cooley2006]_` and `[Neves2011]_`.

PR #336, opened 2023-12-30
Add a global variable to set number of parallel threads

LSchueler:
As pointed out by [pavlovc2](https://github.com/pavlovc2) in discussion #333, it would be helpful to set the number of parallel threads in GSTools directly and not only rely on environment variables.

I didn't have a lot of time to put this together, but in this first draft, you can set the number of threads with the global variable `config.NUM_THREADS`. At the moment, this only works with the Cython code and not with the Rust code. And also, for now, the default value of threads is 1.

TODO

- [ ] also set the number of threads for the Rust code
- [ ] find out how to set `num_threads=None` in the Cython code

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -98,6 +98,8 @@ For the development version, you can do almost the same:
     export GSTOOLS_BUILD_PARALLEL=1
     pip install git+git://github.com/GeoStat-Framework/GSTools.git@main
 
+The number of parallel threads can be set with the global variable `config.NUM_THREADS`.
+
 **Using experimental GSTools-Core for even more speed**
 
 You can install the optional dependency `GSTools-Core <https://github.com/GeoStat-Framework/GSTools-Core>`_,
--- a/src/gstools/config.py
+++ b/src/gstools/config.py
@@ -4,6 +4,8 @@ GStools subpackage providing global variables.
 .. currentmodule:: gstools.config
 
 """
+NUM_THREADS = 1
+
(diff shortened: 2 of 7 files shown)


LSchueler:
The Rust package [GSTools-core](https://github.com/GeoStat-Framework/GSTools-Core) is now ready for setting the number of parallel threads and I'm quite happy with the implementation.

For Cython, I used a bit of an ugly function to set defaults, without having to use Python variables.

I think the only thing left would be to update the changelog, if the review gets through ;-)

MuellerSeb:
This looks promising. Was also winding my head around the fact, that `prange` uses `None` as default value.. this is idiotic to use a python type here. We should create an issue, that prange takes `0` or `-1` to mimic the `None` behavior.
The function you used is ugly but I see that it's the best option at the moment.

Is it testable how many cores are used during a function call? This could be cool to check if the setting is actually working.

LSchueler:
I mean, we could get the numbers of cores in `config.py` and set it there and use `num_threads=1` as the default values in the Cython and Rust functions. But it feels cleaner to get and set the default values in the Cython and Rust functions.

Regarding the testing, I think it's difficult to accurately test the number of threads used, if we don's simply want to check for the `num_threads` variable we just set.
I checked on my machine and the compute times decrease nearly linearly with the number of threads used (for `num_threads < 13` at least). So, locally it works. But I'm not sure how many threads we can use in Github actions and if we really want to check the run times of different function calls.

Issue #343, opened 2024-04-08
Cython: conditional compilation deprecated


MuellerSeb:
https://github.com/cython/cython/issues/4310

The conditional compilation introduced in #336 are deprecated by cython.
I capped the cython version in #340 for now.

@LSchueler this should be cleaned up at some point.

MuellerSeb:
Also cython_lint fails to read these files.

MuellerSeb:
scikit learn has also solved this issue, but with an openmp helper module: https://github.com/scikit-learn/scikit-learn/blob/0c72e5448b9da3b0d9202aa5a2abbd9bff4c238a/sklearn/utils/_openmp_helpers.pyx#L25

MuellerSeb:
Solved temporarily by #354

PR #354, opened 2024-07-04
Drop deprecated conditional compilation

LSchueler:
As mentioned in #343, our very clean and elegant implementation of the conditional import of OpenMP is deprecated and this is my attend to find a new solution.

### DIFF
--- a/src/gstools/krige/base.py
+++ b/src/gstools/krige/base.py
@@ -244,15 +244,15 @@ class Krige(Field):
         if (
             config.USE_GSTOOLS_CORE and config._GSTOOLS_CORE_AVAIL
         ):  # pylint: disable=W0212
-            self._calc_field_krige = (
-                calc_field_krige_gsc  # pylint: disable=E0606, W0201
+            self._calc_field_krige = (  # pylint: disable=W0201
+                calc_field_krige_gsc  # pylint: disable=E0606
             )
             self._calc_field_krige_and_variance = (  # pylint: disable=W0201
                 calc_field_krige_and_variance_gsc  # pylint: disable=E0606
             )
         else:
-            self._calc_field_krige = (
-                calc_field_krige_c  # pylint: disable=W0201
+            self._calc_field_krige = (  # pylint: disable=W0201
+                calc_field_krige_c
             )
             self._calc_field_krige_and_variance = (  # pylint: disable=W0201
                 calc_field_krige_and_variance_c
--- a/src/gstools/krige/base.py
+++ b/src/gstools/krige/base.py
@@ -23,14 +23,17 @@ from gstools.tools.geometric import rotated_main_axes
 from gstools.tools.misc import eval_func
 from gstools.variogram import vario_estimate
 
-if config.USE_RUST:  # pragma: no cover
+if config._GSTOOLS_CORE_AVAIL:  # pragma: no cover
     # pylint: disable=E0401
-    from gstools_core import calc_field_krige, calc_field_krige_and_variance
-else:
-    from gstools.krige.krigesum import (
-        calc_field_krige,
-        calc_field_krige_and_variance,
+    from gstools_core import (
+        calc_field_krige as calc_field_krige_gsc,
+        calc_field_krige_and_variance as calc_field_krige_and_variance_gsc,
     )
+
+from gstools.krige.krigesum import calc_field_krige as calc_field_krige_c
+from gstools.krige.krigesum import (
+    calc_field_krige_and_variance as calc_field_krige_and_variance_c,
+)
 
 __all__ = ["Krige"]
 
@@ -237,6 +240,16 @@ class Krige(Field):
             the kriging error variance
             (if return_var is True and only_mean is False)
         """
+        if config.USE_GSTOOLS_CORE and config._GSTOOLS_CORE_AVAIL:
+            self._calc_field_krige = calc_field_krige_gsc
+            self._calc_field_krige_and_variance = (
+                calc_field_krige_and_variance_gsc
+            )
+        else:
+            self._calc_field_krige = calc_field_krige_c
+            self._calc_field_krige_and_variance = (
+                calc_field_krige_and_variance_c
+            )
         return_var &= not only_mean  # don't return variance when calc. mean
         fld_cnt = 2 if return_var else 1
         default = self.default_field_names[2] if only_mean else None
@@ -284,11 +297,13 @@ class Krige(Field):
 
     def _summate(self, field, krige_var, c_slice, k_vec, return_var):
         if return_var:  # estimate error variance
-            field[c_slice], krige_var[c_slice] = calc_field_krige_and_variance(
-                self._krige_mat, k_vec, self._krige_cond
+            field[c_slice], krige_var[c_slice] = (
+                self._calc_field_krige_and_variance(
+                    self._krige_mat, k_vec, self._krige_cond
+                )
             )
         else:  # solely calculate the interpolated field
-            field[c_slice] = calc_field_krige(
+            field[c_slice] = self._calc_field_krige(
                 self._krige_mat, k_vec, self._krige_cond
             )
 
--- a/src/gstools/krige/base.py
+++ b/src/gstools/krige/base.py
@@ -18,22 +18,23 @@ from scipy.spatial.distance import cdist
 
 from gstools import config
 from gstools.field.base import Field
+from gstools.krige.krigesum import calc_field_krige as calc_field_krige_c
+from gstools.krige.krigesum import (
+    calc_field_krige_and_variance as calc_field_krige_and_variance_c,
+)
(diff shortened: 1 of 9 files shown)


LSchueler:
@MuellerSeb What is this?! Have you ever encountered such an error? - The problems seems to be the `bint` type, not the `const` keyword, when passing bools to Cython. I haven't found anything about this error online.

`ValueError: Buffer dtype mismatch, expected 'const bool' but got 'bool'`

MuellerSeb:
Cool that this is working now.

What I don't like is, that we now have a lot of logic and pylint stuff everywhere we call a compiled algorithm..

What do you think about putting all cython code in a `lib` folder and provide a convenience layer there to provide the right implementation based on `USE_RUST`? We could have some lazy loading logic there as well.

Here is an idea:

```python
import importlib.util

# Step 1: Check if the Rust backend is available (this could be in gstools.config)
RUST_BACKEND_AVAILABLE = importlib.util.find_spec("rust_backend") is not None

# Step 2: Global variable for switching implementations
USE_RUST = RUST_BACKEND_AVAILABLE

# Step 3: Variables to store the implementations
implementations = {
    "rust": {},
    "cython": {}
}


def load_algorithm(func_name):
    if USE_RUST:
        if RUST_BACKEND_AVAILABLE:
            if func_name not in implementations["rust"]:
                import rust_backend
                implementations["rust"][func_name] = getattr(rust_backend, func_name)
        else:
            raise ImportError("Rust backend is not available.")
    else:
        if func_name not in implementations["cython"]:
            import cython_backend
            implementations["cython"][func_name] = getattr(cython_backend, func_name)


def get_algorithm_function(func_name):
    load_algorithm(func_name)
    if USE_RUST and RUST_BACKEND_AVAILABLE and func_name in implementations["rust"]:
        return implementations["rust"][func_name]
    else:
        return implementations["cython"][func_name]


# Step 4: Factory function to create wrapper functions
def create_wrapper_function(func_name):
    def wrapper(*args, **kwargs):
        func = get_algorithm_function(func_name)
        return func(*args, **kwargs)
    return wrapper


# Step 5: Define the list of function names and create wrappers dynamically
# this could also be explicitly done
function_names = ["algorithm", "other_function"]
for name in function_names:
    globals()[name] = create_wrapper_function(name)


# Usage example
USE_RUST = True  # Use Rust implementation if available
result = algorithm(some_arg)
other_result = other_function(another_arg)

USE_RUST = False  # Switch to Cython implementation
result = algorithm(some_arg)
other_result = other_function(another_arg)

USE_RUST = True  # Switch back to Rust implementation
result = algorithm(some_arg)
other_result = other_function(another_arg)
```

Then we can use the algorithms by simply calling them and the backend is selected automatically. Also the algorithm is only loaded if called and switching between rust and cython is easy. We could then also have algorithms that are not implemented in the rust backend and raise an error only for this routine.

Only downside is maybe that the error for the missing rust backend is only thrown with the first function call after USE_RUST was set to true.

LSchueler:
I completely agree that the code looks rather messy right now. But if we really go down that road, I see one problem and one thing to discuss.

Originally, we decided to put the Cython files into the folders of their respective submodules because logically, they belong there. With your approach, we would have the 10 or so "algorithm imports" globally in GSTools. That's not a very clean solution, albeit, some of the if's and pylint things we have in this PR right now would vanish.

The other thing to discuss is that if we actually put all the Cython files into a separate folder, why not create a separate Python package with only the Cython stuff in it, exactly like the GSTools-Core package with the Rust bindings? That would actually make a lot of things much easier for the GSTools package, as the Cython code only changes very seldomly and we could skip all the compilation steps and the rather complicated setup.

MuellerSeb:
> The other thing to discuss is that if we actually put all the Cython files into a separate folder, why not create a separate Python package with only the Cython stuff in it, exactly like the GSTools-Core package with the Rust bindings? That would actually make a lot of things much easier for the GSTools package, as the Cython code only changes very seldomly and we could skip all the compilation steps and the rather complicated setup.

I love the idea of a `GSTools_cython` package. Will try this by forking GSTools and removing everything but the cython code to keep a history.

Your other proposal seems reasonable for the time being.

LSchueler:
awesome!

I just moved all the if-logic into separate functions on the sub-module level. A few things to discuss:

- How detailed should the docstrings of the wrappers be?
- Should the functions start with an underscore? - And then remove the notes from the docstrings.
- What should the default value of the optional argument `num_threads` be? In Cython and Rust, it's `None`. I proposed `num_threads=config.NUM_THREADS` in the commit.

MuellerSeb:
>     * How detailed should the docstrings of the wrappers be?

Wouldn't add doc strings in the wrapper

>     * Should the functions start with an underscore? - And then remove the notes from the docstrings.

Yes, keep it private (and no need for doc strings)
>     * What should the default value of the optional argument `num_threads` be? In Cython and Rust, it's `None`. I proposed `num_threads=config.NUM_THREADS` in the commit.

Would use None in the wrapper like in the cython/rust implementations. Later `num_threads=config.NUM_THREADS` is passed either way.


For new cython backend see here: https://github.com/GeoStat-Framework/GSTools-Cython/pull/1
:tada:

LSchueler:
Okay, I updated the wrappers. But just for code documentation, I included super-short docstrings. I could imagine that those functions would be rather confusing, if they have no description at all.

If you are fine with that decision, LGTM?

src/gstools/variogram/estimator.pyx
```diff
@@ -342,7 +345,7 @@ def structured(
 
 def ma_structured(
     const double[:, :] f,
-    const bint[:, :] mask,
+    uint8[:, :] mask,  # numpy's bools are 8bit vars
```
MuellerSeb, line 348: But we passed the masks as np.int32 before. Was that not working anymore?

src/gstools/variogram/variogram.py
```diff
@@ -471,7 +559,10 @@ def vario_estimate_axis(
...
             field.mask = np.logical_or(field.mask, missing_mask)
         mask = np.ma.getmaskarray(field)
-        if not config.USE_RUST:
+        if (
+            not config.USE_GSTOOLS_CORE
+            and config._GSTOOLS_CORE_AVAIL  # pylint: disable=W0212
+        ):  # pylint: disable=W0212
             mask = np.asarray(mask, dtype=np.int32)
     else:
         field = np.atleast_1d(np.asarray(field, dtype=np.double))
```
MuellerSeb, line 568: Ahh, that is a compatibility issue. Now I get it. So we use bool in both cases now? Can this be removed then?!

src/gstools/krige/base.py
```diff
@@ -39,6 +43,40 @@
 """dict: Standard pseudo-inverse routines"""
 
 
+def _calc_field_krige(krig_mat, krig_vecs, cond, num_threads=None):
```
MuellerSeb, line 46: There is an issue in krigesum.pyx L63, where `num_threads` is defined as follows:
```python
const int num_threads=1
```
This should be just
```python
num_threads=None
```

MuellerSeb (changes requested):
Minor things to clarify. But for a temporary fix, this looks good. Looking forward to put these ugly function selectors in a lib folder.

src/gstools/variogram/estimator.pyx
LSchueler: I have absolutely no clue! But it seems, if we want to use numpy's bool types, `unit8` is the right type to use in Cython. The only thing that I changed that breaks the old code is that we now use function pointers in Python. But don't ask me how that could explain anything ;-)

src/gstools/variogram/variogram.py
LSchueler: Indeed, good eye ;-)

src/gstools/krige/base.py
LSchueler: Also a great catch, thanks!

LSchueler:
I think this is finally ready?!

src/gstools/variogram/estimator.pyx
MuellerSeb: I guess the problem is your if condition for converting the array: `not config.USE_GSTOOLS_CORE and config._GSTOOLS_CORE_AVAIL`

This should be only `not config.USE_GSTOOLS_CORE` since if GSTools-Core is not available (and we obviously want to use cython) this will not evaluate to true and the conversion for cython is not done.

src/gstools/variogram/variogram.py
MuellerSeb: Since we now use the correct type for numpy bools, this conversion also should not happen.

LSchueler: Yeah, in the newest commits, I dropped those lines. I see that that was not clear from my previous message ;-)

MuellerSeb (approved):
LGTM

PR #358, opened 2024-07-15
Remove duplicate readme text from docs' index.rst

LSchueler:
I just saw this in GSTools-Cython and had to try this out in GSTools. Duplicating the readme so many times was pretty annoying.


MuellerSeb:
I only didn't do this because the links to the tutorials are not correctly resolved if you don't look at the stable. I didn't know how to resolve this. A cool thing would be to be able to replace strings (link prefixes) in the markdown file when including it in the rst file.

Maybe we just remove the references to the tutorials? Don't know.. Maybe the the MyST parser provides something. Used it in other projects to include markdown files in sphinx doc.

MuellerSeb:
See here for myst config options: https://myst-parser.readthedocs.io/en/v0.17.1/sphinx/reference.html

MuellerSeb:
I think we could preprocess the readme in the conf.py file and simply remove the `https://geostat-framework.readthedocs.io/projects/gstools/en/stable/` string in every place.

LSchueler:
I was just so excited, because at first glance, it looked as if this was working. I'll close this PR for now, as I think we both have enough other things to do at the moment.

PR #359, opened 2024-07-18
Fix coveralls

LSchueler:
We are having problems with coveralls.io at the moment. By only running coveralls once during the `source_check` job, things should be running fine again.


.github/workflows/main.yml
```diff
@@ -31,6 +31,7 @@ jobs:
         run: |
           python -m pip install --upgrade pip
           pip install -v --editable .[lint]
```
MuellerSeb, line 33:
```suggestion
          pip install -v --editable .[lint, test]
```

MuellerSeb: To push coverage we also need to run tests, right?

.github/workflows/main.yml
```diff
@@ -48,9 +49,15 @@ jobs:
...
-      #- name: cython-lint check
-        #run: |
-          #cython-lint src/gstools/
+      - name: cython-lint check
+        run: |
+          cython-lint src/gstools/
+
+      - name: coveralls check
+        run: |
+          python -m coveralls --service=github
```
MuellerSeb, line 58:
```suggestion
          python -m pytest --cov gstools --cov-report term-missing -v tests/
          python -m coveralls --service=github
```

MuellerSeb: I think we need to run pytest also.

src/gstools/field/summator.pyx
```diff
@@ -10,7 +10,7 @@ if OPENMP:
     cimport openmp
 
 cimport numpy as np
-from libc.math cimport cos, pi, sin, sqrt
+from libc.math cimport cos, sin
```
MuellerSeb, line 13: Ah, linting helps

MuellerSeb (changes requested):
I think we need to run pytest as well.

Also, why is the index/readme hack in here?

.github/workflows/main.yml
LSchueler: True, thanks! I forgot how coveralls.io works...

LSchueler: The syntax seems a bit strange, if you want to install multiple extra dependencies.

src/gstools/field/summator.pyx
LSchueler: It does!

.github/workflows/main.yml
```diff
@@ -30,7 +30,8 @@ jobs:
       - name: Install dependencies
         run: |
           python -m pip install --upgrade pip
-          pip install -v --editable .[lint]
+          pip install -v --editable '.[lint, test]'
```
MuellerSeb, line 33:
```suggestion
          pip install -v --editable .[lint,test]
```

MuellerSeb: I guess the whitespace is the issue

LSchueler:
I think this is ready to be merged.

MuellerSeb:
Please commit my suggestion, then its ready from my side.

.github/workflows/main.yml
LSchueler: I checked again and the syntax causes problems in some shells, e.g. zsh. where `pip install -v --editable .[lint,test]` also doesn't work without the whitespace. I think I don't like whitespace dependent shell commands :-) So, I would just leave the apostrophes, but I can get rid of the whitespace.

PR #364, opened 2024-08-11
Add sum model

MuellerSeb:
This PR adds a `SumModel` class to represent sums of Covariance models.

## Sum-Model
- added `SumModel` class
  - represents sum of covariance models
  - behaves just as a normal covariance model with kriging and field generation
  - covariance models can be added with overloaded `+` operator: `model = m1 + m2`
  - class is subscriptable to access sub-models by index: `m1 = model[0]`
  - included models will get a nugget of 0 and the nugget is stored separately in the sum-model
  - model variance is the sum of the sub-model variances
  - model length-scale is weighted sum of sub-model len-scales, where the weights are the ratios of the sub-models variance to the sum variance (motivated by the integral scale, which satisfies this relation)
  - anisotropy and rotation need to be the same for all included sub-models
  - parameters of the sub-models can be accessed by name with added index suffix: `model[0].nu == model.nu_0`
  - fitting: if `len_scale` is fixed, none of the `len_scale_<i>` can be fixed since len_scale is calculated from variance ratios 
- added Nugget class (empty SumModel)
  - allow len scale of 0 in CovModel to enable a pure nugget model
  - added `zero_var` and `model` attributes to Generator ABC to shortcut field generation for pure nugget models

```python
import gstools as gs
m1 = gs.Gaussian(dim=2, var=1.0, len_scale=2.0)
m2 = gs.Gaussian(dim=2, var=2.0, len_scale=20.0)
model = m1 + m2
model.plot()
```

## Other changes

- removed `var_raw` attribute from CovModel (was rarely used and only relevant for the truncated power law models)
  - BREAKING CHANGE (but not to many should be affected)
  - TPLCovModel now has a `intensity` attribute which calculates what `var_raw` was before
- simplified variogram fitting (`var_raw` was a bad idea in the first place)
- variogram plotting now handles a len-scale of 0 (to properly plot nugget models)
- fitting: when sill is given and var and nugget are deselected from fitting, an error is raised if given var+nugget is not equal to sill (before, they were reset under the hood in a strange way)


MuellerSeb:

LSchueler:
Awesome, I'll need some time to thoroughly go throw your changes.
But I already saw that you claimed example number 11 for the sum model after I did the same for the plurigaussian fields :-D Do we just keep adding new examples in an ascending order according to the time a PR is merged?

MuellerSeb:
First come, first serve :-D

src/gstools/covmodel/sum_tools.py
```diff
@@ -0,0 +1,256 @@
...
+        if np.isclose(summod.ratios[i], 0):
+            skip.add(i)
+        else:
+            wei.append(weights[j])
+        j += 1
+    weights = wei
+    len_sum = sum(summod[i].len_scale * summod.ratios[i] for i in skip)
+    len_diff = len_scale - len_sum
+    if len_diff < 0:
+        msg = "SumModel.set_len_weights: skipped length scales to big."
```
LSchueler, line 196: Please fix grammar

src/gstools/covmodel/fit.py
```diff
@@ -213,70 +213,185 @@ def fit_variogram(
...
     para.update(para_select)
     # check if anisotropy should be fitted or set
     if not isinstance(anis, bool):
         model.anis = anis
         anis = False
-    return para, sill, constrain_sill, anis
+    return para, sill, anis, sum_cfg
+
+
+def _check_sill(model, para_select, sill, bnd, sum_cfg):
```
LSchueler, line 254: Please add a short description of what exactly is checked.

src/gstools/covmodel/fit.py
```diff
@@ -460,36 +622,56 @@ def curve(x, arg1, *args):
...
+                for i, val in zip(sum_cfg["len_fit"], len_vals):
+                    setattr(model, f"len_scale_{i}", val)
+        for i in range(model.size):
+            fit_para[f"var_{i}"] = model.vars[i]
+            fit_para[f"len_scale_{i}"] = model.len_scales[i]
+    # handle sill
+    if sill is not None and para["var"]:
+        nugget = sill - model.var
+        fit_para["nugget"] = nugget
+        model.nugget = nugget
```
LSchueler, line 670: There are quite a few pretty large `if is_sum` branches all over the place here. Maybe you can find a better structure for the code in this module?

LSchueler (changes requested):
Sorry for taking so long to finish this review, but as you know, I had the best reasons ;-)

I'm really looking forward to sum models being integrated into GSTools, but I found quite a few things which should be addressed before merging.

examples/11_sum_model/01_fitting_sum_model.py
```diff
@@ -0,0 +1,76 @@
...
+#
+# A sum model can also be created by a list of model classes together with
+# the common arguments (like dim in this case).
+
+fit_model = gs.SumModel(gs.Spherical, gs.Gaussian, dim=2)
+
+###############################################################################
+# We fit the sum model to the empirical variogram using GSTools' built-in
+# fitting capabilities. As seen in the representation, the variances and length
+# scales of the individual models can be accessed by the attributes
```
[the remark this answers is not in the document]
MuellerSeb: But the variances and length scales are handled differently, since they are stored as a list defining the ratios between the models. So I would keep it like this.

src/gstools/covmodel/fit.py
MuellerSeb: I think this is a task for the future to rework the fitting in general. I think a class for a variogram fitter could be helpful to not pass around all the the information over and over again.

For now I would keep it like this.

MuellerSeb: See #382

MuellerSeb:
@LSchueler fixed (almost) all your complains. Second look? :upside_down_face:

tests/test_sum_model.py
LSchueler, line 1: :-D This diff looks pretty funny. The easiest refactoring ever?! ;-)

MuellerSeb: Jepp :joy:

MuellerSeb:
# LGTM :tada:

PR #370, opened 2024-11-19
Plurigaussian fields

LSchueler:
Plurigaussian simulations (PGS) are a great way to easily increase the flexibility of Gaussian random fields. With this PR I want to directly incorporate them into GSTools. It not only includes the implementation, but also unittests and a few examples, with which users can familiarize themselves with PGS, as I do not find them very intuitive at first.

However, I still have a few open questions:

- Should `PGS` really be a class? I mostly did this for GSTools to keep a more or less constant structure. But I don't see any benefits of using a class here. It only makes it a bit more complicated to use PGSs, as you have to first create an instance and then call it, instead of only calling a `pgs` function. Any opinion on this point @MuellerSeb ?

- At one point we might have to think about a new structure for the examples. In `01_random_field` we have examples for general field generation, but based on the randomization method and some examples showing the Fourier method. Then we have `04_vector_field` with the vector field generation. And now we have `11_plurigaussian`. What do you think @MuellerSeb , should we discuss this soon?

- Is the argument `facies` the best name for a method which is used in many different fields?

- Any other paper(s) we should cite?


LSchueler:
Wow, that's a [curious details](https://stackoverflow.com/a/78552053/2609994) I stumbled upon...

And for some reason, the old `numpy.AxisError` didn't work for me with numpy 2.x, which was moved to `numpy.exceptions.AxisError`, but for backwards compatibility was supposed to be kept in the old place too.

src/gstools/field/pgs.py
```diff
@@ -0,0 +1,108 @@
...
+        Returns
+        -------
+        pgs : :class:`numpy.ndarray`
+            the plurigaussian field
+        """
+        # very clunky way of supporting both np 1.x and 2.x exceptions
+        try:
+            np.AxisError = np.exceptions.AxisError
+        except AttributeError:
+            ...
```
MuellerSeb, line 77: I did something similar for emcee. I would import this at the top: https://github.com/dfm/emcee/pull/517/files#diff-1697674b19133cd11b2fbdf0037fcac453954401163ba710c3c0788fa19402f4

LSchueler: I moved the imports to the top.

LSchueler:
Thanks to @EJRicketts's feedback, I updated the examples and fixed a bug, where the L field could have an offset.

EJRicketts:
Thank you or sharing this with me, it's nice to see that the offset has been sorted :) This was my only major comment.

In general, I think it would be nice to include an example that shows the use of conditional random fields and also periodic random fields. 

For field scale problems with associated experimental observations, combining a well chosen lithotype with conditional random fields can be quite powerful in getting nice realistic representations. 

With respect to the periodic case, periodicity in the input fields will result in periodicity in the final field, which for my research purposes have been useful in material characterisation. I published some work on this for cementitious materials recently: https://doi.org/10.1007/s11242-024-02074-z
It would be easy to replicate these results for the example, but perhaps a more geoenvironmental example would be best! In any case, a small mention of the paper in the example would be much appreciated 😊

Regarding citing works, there is a nice book on PGS: https://link.springer.com/book/10.1007/978-3-642-19607-2
The authors are (mostly) from the birth place of PGS (Mines-Paristech), some of which are the original authors who took the idea of truncated random fields and extended it to PGS. It feels important to include this.
Xavier Emery has also had quite a big influence on PGS, so it might be worth citing their work also: https://doi.org/10.1016/j.cageo.2007.01.006

LSchueler:
@MuellerSeb This is ready for review

examples/11_plurigaussian/04_3d_pgs.py
```diff
@@ -0,0 +1,82 @@
...
+#    running the example yourself. You will get an interactive version of this
+#    screenshot.
+
+# grid.contour(isosurfaces=8).plot()
+
+###############################################################################
+#
+# .. image:: https://github.com/GeoStat-Framework/GeoStat-Framework.github.io/raw/master/img/3d_pgs.png
+#    :width: 400px
+#    :align: center
```
MuellerSeb, line 82: This is not shown in the docs.

examples/11_plurigaussian/02_spatial_relations.py
```diff
@@ -0,0 +1,111 @@
...
+        + R[1]
+        + 3
+        + 2 * i : M[1] // 2
+        + S1[1]
+        + R[1]
+        + 4
+        + 2 * i,
+    ] = (
+        4 + i
+    )
```
MuellerSeb, line 83: I don't know if this is common in the literature about PGS, but I really dislike all the single uppercase Letters for matrices used in this context.

I think the whole API and documentation of GSTools is quiet descriptive, or at least tries to be, and this reads almost like Fortran 77 code.

Couldn't we come up with something like a descriptive and self explaining naming convention for these "things"?

src/gstools/field/pgs.py
```diff
@@ -0,0 +1,173 @@
...
+        ----------
+        L : :class:`numpy.ndarray`
+            A `dim` dimensional structured field, whose values are mapped to the PGS.
+            It does not have to have the same shape as the `fields`, as the indices are
+            automatically scaled.
+        Returns
+        -------
+        pgs : :class:`numpy.ndarray`
+            the plurigaussian field
+        """
```
MuellerSeb, line 94: Here the same applies. I wouldn't use `L` for this argument. It should be something concise and understandable.

src/gstools/field/pgs.py
```diff
@@ -0,0 +1,173 @@
...
+
+        P_dig = []
+        for d in range(self._dim):
+            P_dig.append(np.digitize(self._mapping[:, d], self._pos_l[d]))
+
+        # once Py3.11 has reached its EOL, we can drop the 1-tuple :-)
+        return self._L[(*P_dig,)]
+
+    def calc_L_axes(self, L_shape):
+        """Calculate the axes on which the L field is defined.
```
MuellerSeb, line 108: `L` everywhere...

MuellerSeb (changes requested):
This is awesome. And so cool to see what can be done in geo-statistics with some sugar on it.

A well done contribution and beautiful examples. :tada:
I would only like to see better naming for the used variables and especially for method arguments and attributes. Maybe this is a common thing in the pgs community, but I would like to have descriptive names in GSTools.

What do you think @LSchueler ?

examples/11_plurigaussian/04_3d_pgs.py
LSchueler: I *think*, I *hope* this is shown, as soon as this branch is merged into main. At least that's how I remember I did it the last time, I uploaded an image to the docs.
Can you confirm this?

examples/11_plurigaussian/02_spatial_relations.py
LSchueler: Ouch, that Fortran 77 comparison hurt my feelings :-D

I found that code pretty ugly already and didn't want to reduce readability by introducing longer variable names. But changed most of the variable names to longer and more descriptive ones, because your criticism is absolutely valid. I didn't change all of them and kept e.g. `M`, similarly as we often use `N` for the number of grid points.
Are you okay with this compromise?

LSchueler:
If you think that the image will be shown once this is merged and you like my new variable names, this could be merged?

examples/11_plurigaussian/04_3d_pgs.py
MuellerSeb: Actually, we shouldn't use html adresses for images stored in docs folder (as you can see by the use of "master"). We did that in the past due to stupidity. Sphinx should be capable of using local paths to reference images.

PR #376, opened 2025-02-04
Outsource Cython backend

MuellerSeb:
Now that we have [GSTools-Cython](https://github.com/GeoStat-Framework/GSTools-Cython), we can remove the cython code from GSTools.


.github/workflows/main.yml
```diff
@@ -123,8 +95,9 @@ jobs:
           pip install build
 
       - name: Install GSTools
-        env:
-          GSTOOLS_BUILD_PARALLEL: 1
+        # env:
+        #   GSTOOLS_BUILD_PARALLEL: 1
+        #   pip install --no-binary gstools-cython -v --editable .[test]
```
MuellerSeb, line 100: Not sure about this. Should we test with parallel @LSchueler?
In the GSTools-Cython repo, we already test the cython routines with parallel execution.

LSchueler: Two things come to mind.
- Depending on what kind of CI/CD resources we get from Github, the tests might finish quicker.
- Some time ago, we had very rare bugs, which where probably some kind of race conditions, only happening with parallel code execution. In case this ever happens again, we might be able to catch them earlier, if we test more in parallel.

But I'm also not sure how much weight you should give those two points.

MuellerSeb: I think it is better to check the real world scenario where a pre-built wheel of gstools-cython is uesd. When rebuilding gstools-cython local, we could miss issues that have far more impact.

And I would argue against doing both. So for now I would be fine with only testing parallel execution in the gstools-cython repo (which should be enough IMHO).

LSchueler (approved):
It's really cool to see how fast GSTools now "build" with GSTools-Cython already built. Thank you this PR and for the new Cython repo!!

Two things to think about.

- Now that I read the installation guide again, I think that the part about building Cython with parallel support sounds pretty complicated, which it really isn't. Maybe first provide the code block and then the explanation for those who are really interested?

- Because the Rust code runs parallel without any extra installation steps when using pip, maybe that installation guide should come before the Cython part in order to keep it as simple as possible for the users? And maybe we should tone the text down a bit. It's probably not going to completely replace the Cython code anytime soon ;-)

But apart from that, I don't see anything to complain about, so LGTM ;-)

MuellerSeb:
I would keep gstools-core on second place in docs, since there is no conda package for gstools-core. ATM I have no experience with putting a rust package on conda-forge. But I think it's no rocket science.

MuellerSeb:
# LGTM :tada:

PR #379, opened 2025-02-08
Add CSV export method to Field class

### LINKED ISSUE #367: [ENH] CSV export method for Field class
Since we store all generated fields by name for the same pos tuple, it would make perfectly sense to provide a CSV export routine, that exports all fields as unstructured list of points and values.

Vector fields, could get suffixes `_0`, `_1`, `_2` and so on.

MuellerSeb:
closes #367 

This new method `csv_export` allows to export all fields of a field class together with the pos tuple into a csv file.
Structured fields will be converted to unstructured to get flat arrays.

Example:

```python
from gstools import SRF, Gaussian

x = y = range(3)
model = Gaussian(dim=2)
srf = SRF(model, seed=20170519)
srf((x, y), mesh_type="structured")
srf.csv_export("fields.csv")
```
Resulting file `fields.csv`:
```csv
x,y,field
0.0,0.0,-0.22138986032350452
0.0,1.0,-0.38477418693294124
0.0,2.0,-0.11175947311945715
1.0,0.0,-0.7538723514717599
1.0,1.0,0.3501084035555686
1.0,2.0,1.62100148329647
2.0,0.0,0.7236152951349311
2.0,1.0,-0.058499764834117715
2.0,2.0,0.3130225131979303
```

### DIFF
--- a/src/gstools/field/base.py
+++ b/src/gstools/field/base.py
@@ -434,6 +434,72 @@ class Field:
             fieldname=fieldname,
         )
 
+    def csv_export(self, filename, fields=None):  # pragma: no cover
+        """Export the stored field(s) to csv with unstructured mesh type.
+
+        First columns with be the components of the pos tuple followed by
+        the selected fields.
+
+        Parameters
+        ----------
+        filename : :class:`str`
+            Filename of the file to be saved, including the path.
+        fields : :class:`list` of :class:`str`, optional
+            Fields that should be stored. All by default.
+        """
+        fields = self.field_names if fields is None else fields
+        if not fields:
+            return
+        if not all(field in self for field in fields):
+            msg = f"csv_export: some fields are unknown: {fields}"
+            raise ValueError(msg)
+        # generate axis names
+        s_dim = self.dim - int(self.temporal)
+        if self.latlon:
+            p_names = ["lat", "lon"]
+        elif self.dim < 4:
+            p_names = ["x", "y", "z"][:s_dim]
+        else:
+            p_names = [f"x{d}" for d in range(s_dim)]
+        if self.temporal:
+            p_names.append("t")
+        # generate fields
+        if self.mesh_type != "unstructured":
+            pos = generate_grid(self.pos)
+            if self.value_type == "vector":
+                out_f = {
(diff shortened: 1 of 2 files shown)


LSchueler:
Hey hey,

thank you for the initiative!
Originally, we decided against implementing import and export support, primarily because was that Python has many great tools supporting import and export. Those tools can effectively be used as pre- and postprocessing options for GSTools.
However, at least supporting rudimentary import and export methods can be very convenient. That said, everything else beyond rudimentary support would either amount in a lot of work or large dependencies like Polars or Pandas.
Additionally, CSV is a format which shouldn't receive even more standalone support than it has already, as summarized poignantly [here](https://kaveland.no/posts/2024-03-10-friends-dont-let-friends-export-csv/). So, how about this compromise: We support export functionality through the main Numpy export functions, like
- [savetxt](https://numpy.org/doc/stable/reference/generated/numpy.savetxt.html)
- [save](https://numpy.org/doc/stable/reference/generated/numpy.save.html)
- [savez](https://numpy.org/doc/stable/reference/generated/numpy.savez.html)
- [savez_compressed](https://numpy.org/doc/stable/reference/generated/numpy.savez_compressed.html)

I think this should be easily doable, by providing another argument with which you can choose your export function/ format. I think it would even be okay for me, if CSV stays the default format.
What do you think?

LSchueler:
Yes, I had a look at your implementation. But from a quick check, it seems that `savetxt`, `save`, ... all look very similar, they also take variable names, ...
You can even pickle your variables ;-)

I like the name of the function and I think you should keep `csv_export`.

Why would you want to specifically support json export? I don't think that json is a very fitting format for GSTools and would much more prefer other formats.

MuellerSeb:
But json for example to store CovModel configuration? I think this would be really cool.

MuellerSeb:
I think all the others export routines are more meant to be used for storing and loading again, since they are not human readable file formats.
I don't want to implement a "load-from-file" method at the moment and as you said, the users could use the Python eco-system to create files as they want.

LSchueler:
> But json for example to store CovModel configuration? I think this would be really cool.

We should probably shift this discussion to another time ;-)

PR #386, opened 2025-06-17
Drop Python 3.8, depend on scipy>=1.6.0

mwtoews:
Python 3.8 reached it's end-of-life in 2024-10-07 ([PEP 569](https://peps.python.org/pep-0569/)), so should be dropped.

Change SciPy dependency to >=[1.6.0](https://pypi.org/project/scipy/1.6.0/) to overlap with Python 3.9. Similar with NumPyDoc. All other deps have not changed, as they appear to overlap with Python 3.9.

gstools-core only supports Python>=3.10, so add conditional so that (e.g.) `uv sync --all-extras` succeeds.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,7 +8,7 @@ requires = [
 build-backend = "hatchling.build"
 
 [project]
-requires-python = ">=3.8"
+requires-python = ">=3.9"
 name = "gstools"
 description = "GSTools: A geostatistical toolbox."
 authors = [
@@ -31,7 +31,6 @@ classifiers = [
     "Programming Language :: Python",
     "Programming Language :: Python :: 3",
     "Programming Language :: Python :: 3 :: Only",
-    "Programming Language :: Python :: 3.8",
     "Programming Language :: Python :: 3.9",
     "Programming Language :: Python :: 3.10",
     "Programming Language :: Python :: 3.11",
@@ -51,7 +50,7 @@ dependencies = [
     "meshio>=5.1.0",
     "numpy>=1.20.0",
     "pyevtk>=1.1.1",
-    "scipy>=1.1.0",
+    "scipy>=1.6.0",
 ]
 
 [project.optional-dependencies]
@@ -59,7 +58,7 @@ doc = [
     "myst_parser",
     "matplotlib>=3.7",
     "meshzoo>=0.7",
-    "numpydoc>=1.1",
+    "numpydoc>=1.2",
     "pykrige>=1.5,<2",
     "pyvista>=0.40",
     "sphinx>=7",
@@ -71,7 +70,7 @@ plotting = [
     "matplotlib>=3.7",
     "pyvista>=0.40",
 ]
-rust = ["gstools_core>=1.0.0"]
+rust = ["gstools_core>=1.0.0; python_version >= '3.10'"]
 test = ["pytest-cov>=3"]
 lint = [
     "black>=24",
@@ -115,7 +114,6 @@ line_length = 79
 [tool.black]
 line-length = 79
 target-version = [
-    "py38",
     "py39",
     "py310",
     "py311",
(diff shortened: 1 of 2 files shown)


mwtoews:
Extending support 3.8 is fine in principle, but the current setup is broken for tools like [`uv sync`](https://docs.astral.sh/uv/reference/cli/#uv-sync), which raises errors when dependencies are not met, e.g. I see:
```console
$ uv sync
  × No solution found when resolving dependencies for split
  │ (python_full_version == '3.8.*'):
  ╰─▶ Because the requested Python version (>=3.8) does not satisfy
      Python>=3.9 and all versions of stonefish-license-manager
      depend on Python>=3.9, we can conclude that all versions of
      stonefish-license-manager cannot be used.
      And because meshzoo==0.11.6 depends on stonefish-license-manager and
      only meshzoo==0.11.6 is available, we can conclude that meshzoo==0.11.6
      cannot be used.
      And because gstools[doc] depends on meshzoo==0.11.6 and your project
      requires gstools[doc], we can conclude that your project's requirements
      are unsatisfiable.

      hint: The `requires-python` value (>=3.8) includes Python versions
      that are not supported by your dependencies (e.g., all versions of
      stonefish-license-manager only supports >=3.9). Consider using a more
      restrictive `requires-python` value (like >=3.9).
```
In this case, a minimal patch to resolve the dependency issue is:
```diff
diff --git a/pyproject.toml b/pyproject.toml
index 3e09aaf..df2497b 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -58,7 +58,7 @@ dependencies = [
 doc = [
     "myst_parser",
     "matplotlib>=3.7",
-    "meshzoo>=0.7",
+    "meshzoo>=0.7; python_version >= '3.9'",
     "numpydoc>=1.1",
     "pykrige>=1.5,<2",
     "pyvista>=0.40",
@@ -71,7 +71,7 @@ plotting = [
     "matplotlib>=3.7",
     "pyvista>=0.40",
 ]
-rust = ["gstools_core>=1.0.0"]
+rust = ["gstools_core>=1.0.0; python_version >= '3.10'"]
 test = ["pytest-cov>=3"]
 lint = [
     "black>=24",
```
(In this case it looks like [meshzoo has only one release in PyPi](https://pypi.org/project/meshzoo/#history), which depends on other projects that require Python 3.9).

The SciPy dep was only incremented to overlap Python 3.9 support, since SciPy 1.5.4 only supported up to Python 3.8.

But as to the broader question of dropping Python beyond it's intended EOL, there is a spectrum of opinions. Some say to relax it to as many versions as possible, because the code still functions the same. But this requires more testing and more developer time to maintain in a complex dependency ecosystem (e.g. meshzoo). And the other side of the spectrum is [SPEC 0](https://scientific-python.org/specs/spec-0000/) that suggests supporting the previous ~3 releases (currently Python 3.11-3.13), as adopted by NumPy. This is is usually the approach for projects that compile many variations of binary wheels, to conserve computer time and disk space.

So if you're certain folks out there are actually using Python 3.8, then there is no issue in keeping it in support, as this is a pure Python project with a single wheel. I can re-adjust this PR to suite, let me know.

LSchueler:
Thanks for raising this issue!
I don't really see a problem in dropping py3.8 support for the optional dependency. I guess not many people with py3.8, if any, will be compiling the docs? But yeah I agree that the meshzoo dependency has become the fundamental problem here.

I'm going to check the dependencies of gstools-core next week. And I hope to also find some time to drop the meshzoo dependency.

pyproject.toml
```diff
@@ -71,7 +70,7 @@ plotting = [
     "matplotlib>=3.7",
     "pyvista>=0.40",
 ]
-rust = ["gstools_core>=1.0.0"]
+rust = ["gstools_core>=1.0.0; python_version >= '3.10'"]
```
MuellerSeb, line 73: @LSchueler maybe we should integrate at least this?

LSchueler:
We'll keep supporting Py3.8 for now, the meshzoo mess was fixed by dropping the dependency completely, and gstools-core now supports Py3.8 - Py3.14.

PR #387, opened 2025-06-27
Update PGS to contain tree based lithotype rules

EJRicketts:
When using spatial lithotypes, we often fall into the trap of restricting ourself to using the same number of input spatial random fields as the dimensions of the simulation domain. PGS as a method does not require this, so it would be nice to be able to use an arbitrary number of random fields. This is what the decision tree type implementation allows for. 

Much of the PGS class has been refactored, and some new examples have been added to highlight how to use the tree based lithotype architectures.

Questions:
- Would it be worth adding the `ellipse()` and `ellipsoid()` functions as helper functions, or just allow people to define their own? Other shapes could also be added
- Are we happy with the `DecisionTree` class being a child of `PGS`? I thought to add as a separate class in `pgs.py`, but having it as a child seemed cleaner


EJRicketts:
This is the beauty of having a bit more free time! 

I have gone through and checked through with `black`, `isort`, and `pylint` and things look all good! `pylint` does throw some warnings for `Module gstools.covmodel.tools` and `Module gstools.transform.array`, but I have left these

I also need to see why `test_assertions()` is not working in `TestPGS`, but I'm not sure why as it is using the original PGS implementation

LSchueler:
Hi Evan,

I finally found some time to have a look at your code, sorry for having to wait for so long. Your examples are beautiful!!

Here are a few first remarks.

examples/11_plurigaussian/07_tree_pgs.py
```diff
@@ -0,0 +1,133 @@
+"""
+PGS through decision trees
+--------------------------
+
+In typical PGS workflows, the lithotype is defined through a spatial rule. A
+more flexible approach can be taken such that the lithotype is represented as a
+decision tree. More specifically, this is given as a binary tree, where each
+node is a decision based on the values of the spatial random
+fields [Sektnan et al., 2024](https://doi.org/10.1007/s11004-024-10162-5). The
```
LSchueler: Unfortunately, this doesn't work in sphinx. I just realised that I made the same mistake in my own PGS examples. I'll have to fix those. The correct syntax is

```python
`Sektnan et al., 2024 <https://doi.org/10.1007/s11004-024-10162-5>`_
```

examples/11_plurigaussian/07_tree_pgs.py
```diff
@@ -0,0 +1,133 @@
...
+
+def ellipse(data, key1, key2, c1, c2, s1, s2, angle=0):
+    x, y = data[key1] - c1, data[key2] - c2
+
+    if angle:
+        theta = np.deg2rad(angle)
+        c, s = np.cos(theta), np.sin(theta)
+        x, y = x * c + y * s, -x * s + y * c
+
+    return (x / s1) ** 2 + (y / s2) ** 2 <= 1
```
LSchueler, line 52: I think it would be a great idea to put functions like `ellipse` and `ellipsoid` directly into GSTools' code, probably into `src/gstools/field/tools.py`.
But I could also do that some time after this PR is merged.

examples/11_plurigaussian/08_multi_field_pgs.py
```diff
@@ -0,0 +1,117 @@
+"""
+From bigaussian to plurigaussian simulation
+--------------------------------------------------
+
+In many PGS implementations, the dimensions of the simulation domain often
+matches the number of fields that are supplied. However, this is not a
+requirement of the PGS algorithm. In fact, it is possible to use multiple
+spatial random fields in PGS, which can be useful for more complex lithotype
+definitions. In this example, we will demonstrate how to use multiple SRFs in
+PGS. In GSTools, this is done by utlising the tree based architecture.
```
LSchueler: spelling

examples/11_plurigaussian/README.rst
```diff
@@ -1,14 +1,27 @@
...
+Many PGS approaches constrain the number of input SRFs to match the
+dimensions of the simulation domain. This constraint stems from the
+implementation, not the method itself.
+With a spatial lithotype, we perform bigaussian and trigaussian
+simulations for two- and three-dimensional realisations, respectively.
+In contrast, the tree-based approach allows an arbitrary number of SRFs,
+yielding a fully *pluri*gaussian simulation.
+This may sound more complicated than it is; we will clarify everything
+in the examples that follow.
+
```
LSchueler, line 24: Thank you for this text :-)

LSchueler (changes requested):
So far this looks really good. There are two things which are missing so far.

1. Unit tests are still missing for the new features.

2. The parameters in the docstrings should generate links to their type definitions, e.g. with
```python 
:any:`DecisionNode`
```
You can have a look at the docstrings in e.g. src/gstools/field/srf.py.

But it's really nice that you mention the possible exceptions. One day, we should add that to all docstrings...

I'll need some more time to go through your code in detail, but I really hope to find some more time this week.

examples/11_plurigaussian/07_tree_pgs.py
EJRicketts: No problem. I will leave this for another time :)

examples/11_plurigaussian/README.rst
EJRicketts: 😊

EJRicketts:
All comments have been addressed 😊 The main changes related to adding a full unit test and making sure docstrings would link in the documentation

LSchueler (approved):
Sorry again for the long wait  and thank you so much for this great addition to the PGS functionality. Your decision tree approach generalises my first approach so much!

PR #391, opened 2025-07-09
Replace pylint with ruff

### LINKED ISSUE #390: [Maintanance] use Ruff for linting and formatting
We had several issues with pylint in the past (breaking changes, false-positive errors, long runtimes, e.g. #389)

Now, there is [ruff](https://docs.astral.sh/ruff), which basically is 100x faster then pylint and it can also drop-in replace [black](https://black.readthedocs.io) and [isort](https://pycqa.github.io/isort).

Since it is so well developed and documented, we should think about switching to ruff.

LSchueler:
As mentioned (#390), the issues we have been having with pylint have slowly increased and with the much faster ruff, we have a nice alternative to pylint.

With this PR I have completely replaces pylint with ruff.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -43,7 +43,7 @@ jobs:
 
       - name: ruff format check
         run: |
-          python -m ruff check src/gstools/
+          python -m ruff format --diff src/gstools/
 
       - name: coveralls check
         run: |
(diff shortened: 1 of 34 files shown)


MuellerSeb:
This is great! Thanks for this contribution.

In the CONTRIBUTION.md is a section about the pylint settings that could be replaced with the required ruff checks. Then this is ready for merging.

LSchueler:
Thanks! I interpret your message as an informal holiday review, so LGTM!

MuellerSeb:
Cool. But there is still the whole section about PyLint in the CONTRIBUTING.md file.

PR #392, opened 2025-07-22
Ruff improvements

LSchueler:
Following up on PR #391, here are a few improvements, making it a little bit easier to work with ruff. Also, the CONTRIBUTIONS.md is now up-to-date. I don't think that we have to give a detailed list of our linter settings, as ruff automatically applies the settings we provide in the pyproject.toml file.

Keeping in the spirit of linting and standard formats, I have applied common formatting rules to pyproject.toml too.


MuellerSeb:
I love how ruff just works. LGTM

MuellerSeb (approved):

LSchueler:
Awesome! Yeah about the "it just works" bit... maybe we should have a beer together and think about code styles we like and which of these rules we want to include ;-)
https://docs.astral.sh/ruff/rules/

Issue #402, opened 2026-04-28
Write benchmarks for current DS-MPS impementation


LSchueler:

jeilealr:
### Benchmarking update

I decided to add an ASV benchmark suite for the current GSTools/DS-MPS workflows, plus a small `cProfile` helper to inspect where runtime is spent in the current checkout.

### Why ASV and cProfile?

ASV, [Airspeed Velocity](https://asv.readthedocs.io/en/latest/), is used for repeatable performance benchmarking across commits. It is especially useful when we want to know whether a code change made a workflow faster or slower over time. In this implementation, I also added a backend comparison between the Cython fallback and the Rust backend.

ASV is used by well-known scientific Python projects such as NumPy, SciPy, pandas, scikit-image, and Astropy to track performance regressions across versions.

`cProfile` is complementary. It does not compare commits. Instead, it profiles one run in the current checkout and shows where time is spent inside that workflow. I used it after ASV to understand which internal functions dominate runtime.

A walkthrough of the benchmark implementation is available in the README in my fork: [benchmarks/README.md](https://github.com/jeilealr/GSTools/blob/benchmarking/benchmarks/README.md). I think this is a solid first step to track the current performance status of the repository in a reproducible way.

### Initial results

The implementation produces several ASV plots and terminal summaries. I will share the main results here, with two representative plots.

The benchmark was run across the last five commits on `main`, comparing both backends. The differences between commits are not large, but the backend differences are clear.

### ASV time example: kriging workflow

<img width="1904" height="928" alt="Image" src="https://github.com/user-attachments/assets/5cf549ed-bf8d-491a-9e46-5fe0b0b51414" />

The y-axis is runtime in microseconds, so lower is faster. The x-axis is the git commit history included in the ASV run. Each line is one combination of backend and kriging case:

- `backend=cython_fallback`: GSTools running without the Rust core backend.
- `backend=rust_core`: GSTools running with `gstools_core`.
- `small_30x500`: 30 conditioning points, 500 target prediction points.
- `large_120x2000`: 120 conditioning points, 2,000 target prediction points.
- `extra_large_360x6000`: 360 conditioning points, 6,000 target prediction points.

The lines are mostly flat across the plotted commits, which means the benchmark is stable over the selected history and does not show a strong performance regression in this range.

[...]

PR #410, opened 2026-05-15
Benchmarking

jeilealr:
# Write Benchmarks for Current DS-MPS Implementation

## Summary

This PR adds a reproducible benchmarking setup for GSTools using
[Airspeed Velocity](https://asv.readthedocs.io/) plus a small complementary
`cProfile` helper.

The goal is to establish a measurement baseline before optimizing or extending
the DS-MPS implementation. The benchmarks focus on representative GSTools
workflows rather than isolated micro-functions:

- variogram estimation
- global kriging
- spatial random field generation
- conditioned spatial random field generation

The suite can compare the current Cython fallback backend with the Rust backend
from `gstools_core`, and it can report both runtime and peak-memory results.

## What Changed

- Added the default ASV configuration in `asv.conf.json`.
- Added benchmark workflows in `benchmarks/benchmark_backends.py`.
- Added helper scripts for:
  - Rust-vs-Cython speedup summaries
  - cProfile workflow profiling
  - Cython OpenMP verification
  - macOS OpenMP-enabled `gstools-cython` installation inside ASV
- Added an optional macOS OpenMP ASV configuration in
  `asv.macos-openmp.conf.json`.
- Added a detailed benchmark guide in `benchmarks/README.md`.
- Added the `benchmark` optional dependency extra.
- Added a GitHub Actions workflow for benchmark checks.

Most of the detailed usage instructions, benchmark case explanations, and
interpretation notes are intentionally kept in `benchmarks/README.md` to avoid
duplicating the guide in this PR description.

## Why ASV and cProfile?

ASV is used for repeatable benchmark results across commits. This makes it
useful for detecting performance regressions and tracking how workflows change
over time. It is also the standard benchmarking tool used by several scientific
Python projects, including NumPy, SciPy, pandas, scikit-image, and Astropy.

`cProfile` is used for a different purpose: it profiles one selected workflow in
the current checkout and shows where runtime is spent internally. In this PR,
the cProfile helper imports the same benchmark classes used by ASV, so the
profiling cases stay aligned with the benchmark suite.

## Initial Observations

The initial benchmark runs show that Rust is generally faster for larger
variogram and kriging workloads. For very small cases, Rust can be similar or
slightly slower because fixed overhead dominates. With one thread, Rust is not
always faster for every SRF case, so the backend comparison is workflow- and
problem-size-dependent.

The cProfile results suggest that the slowest parts of the measured workflows
are mostly inside compiled backend/numerical routines. That means Python-level
cleanup is unlikely to be the main optimization path for these cases. The next
useful optimization work should focus on algorithmic cost reduction and backend
implementation details.

## Follow-Up Tasks

- [ ] Test and document the OpenMP benchmark setup on Linux.
- [ ] Test and document the OpenMP benchmark setup on Windows.
- [ ] Test and document the OpenMP benchmark setup on an HPC environment.
- [ ] Decide whether Linux, Windows, and HPC need separate ASV config files or can
  share one generalized OpenMP setup.
- [ ] Connect these benchmarks more directly to the DS-MPS implementation once the
  current migration work is ready to benchmark.


jeilealr:
see #402

asv.conf.json
```diff
@@ -0,0 +1,32 @@
+{
+  "version": 1,
+  "project": "GSTools",
+  "project_url": "https://github.com/jeilealr/GSTools",
```
LSchueler, line 4: Check all the URLs, they are pointing to your fork, not the original repo

benchmarks/tools/profile_benchmark_workflows.py
LSchueler, line 1: This file is huge! I think splitting it into something like a parser and a renderer would make it more readable. I don't know a lot about html etc., but can you outsource the `render_html` stuff into some kind of template?

tests/test_plot_case_backend_comparison.py
LSchueler, line 1: I think it would be helpful to add some tests to check if ASV introduces breaking changes (API or result format) in the future.

benchmarks/benchmark_two_point_statistics.py
```diff
@@ -0,0 +1,368 @@
...
+            kwargs = {"sampling_size": 4500, "sampling_seed": 20220505}
+        with gstools_backend(_use_core(backend), threads):
+            gs.vario_estimate(
+                pos,
+                field,
+                bins,
+                mesh_type="unstructured",
+                return_counts=True,
+                **kwargs,
+            )
```
LSchueler, line 217: I think you should write an internal function like `_run_variogram` which gets called from both `peakmem_variogram_estimate` and `time_variogram_estimate`. The same goes for kriging

.github/workflows/benchmark-readiness.yml
LSchueler, line 1: Do you think this is necessary once your work on the benchmarks is ready? In case it is, can you include the checks into the other actions in order to reduce the CI runtimes?

pyproject.toml
```diff
@@ -83,6 +84,7 @@ fallback_version = "0.0.0.dev0"
 
 [tool.hatch.version.raw-options]
 local_scheme = "no-local-version"
+tag_regex = "^v?(?P<version>[0-9]+\\.[0-9]+.*)$"
```
LSchueler, line 87: Why have you added this?

benchmarks/tools/render_benchmark_comment.py
```diff
@@ -0,0 +1,103 @@
...
+    else:
+        badge = f"✅ No significant changes detected · {note}"
+
+    return (
+        "<!-- gstools-openmp-benchmark -->\n"
+        "## OpenMP Benchmark Results\n\n"
+        f"**Base:** `{base}` → **Head:** `{head}`{report_link}\n\n"
+        f"{badge}\n\n"
+        "<details>\n<summary>Full ASV comparison</summary>\n\n"
+        f"```\n{comparison}\n```\n\n"
```
LSchueler, line 79: Secure this against content injection.

LSchueler (changes requested):
Thanks for your great work! I think once you fixed the requested changes, this is ready to be merged.

asv.conf.json
jeilealr: All jeilealr references are gone. Important note on this topic. There are still two things we need for the benchmarking to work. Unfortunately, I cannot do this myself due to permissions. The two missing steps are: 1) GeoStat-Framework/GSTools-Benchmarks must exist with a main branch and Pages enabled (creating an empty README.md file in main branch and pushing the changes, and then to enable Pages in repo: Settings → Pages → source "Deploy from a branch", branch main, folder / (root)). And 2) BENCHMARKS_REPO_TOKEN secret with push rights must be added to GeoStat-Framework/GSTools (on GeoStat-Framework/GSTools: Settings → Secrets and variables → Actions → New repository secret → name BENCHMARKS_REPO_TOKEN → paste the token).

benchmarks/tools/profile_benchmark_workflows.py
jeilealr: I think you were refering in this comment to file plot_case_backend_comparison.py. I moved the content of the html into a templates folder to have the python script much more clearer. I

benchmarks/benchmark_two_point_statistics.py
jeilealr: Good point. Now _run_variogram, _run_krige, and _run_field for the random-field class; each time_*/peakmem_* pair now delegates to its helper. ASV discovers the same six benchmarks with identical parameters, so existing results stay comparable and we reduced 46 lines :)

pyproject.toml
jeilealr: Must have been there from some time and didn't remove it, I just checked and this is not needed so I removed it.

benchmarks/tools/render_benchmark_comment.py
jeilealr: Done. The comparison text can no longer close its code fence, SHAs are validated as hex, and the report link is only kept if it is an artifact URL of this repository. The workflow_run comment job now takes only the PR number and comparison text from the untrusted artifact, verifies the number against the run's head SHA, and resolves commits and artifact URL from the API. Tests added in tests/test_render_benchmark_comment.py.

tests/test_plot_case_backend_comparison.py
jeilealr: Yes, I agree. I've added tests/test_asv_integration.py, which runs the installed ASV on one fast case and checks the result-JSON layout, asv compare output, asv publish, and that both config files load (runs in source_check, ~7 s).

asv.conf.json
jeilealr: @LSchueler This will be something left to do that I cannot due to permissions

jeilealr:
Hi @LSchueler 

Thank you for having a look at the repo, I addressed all the comments and check that everything is good to go. The only last step I left it in a comment above (first push into benchmark repo and creating BENCHMARKS_REPO_TOKEN)

PR #411, opened 2026-05-28
Drop meshzoo dependency

LSchueler:
With this PR we are dropping the dependency on the by now closed-sourced Python package meshzoo. We only ever needed that package for a small example, so this introduces hardly any changes.


MuellerSeb (approved):
LGTM.

PR #413, opened 2026-05-29
Update gstools-core dependency

LSchueler:
The old versions do not support Py3.14

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -65,7 +65,7 @@ doc = [
     "sphinxcontrib-youtube>=1.1",
 ]
 plotting = ["matplotlib>=3.7", "pyvista>=0.40"]
-rust = ["gstools_core>=1.0.0"]
+rust = ["gstools_core>=1.3.0"]
 test = ["pytest-cov>=3"]
 lint = ["ruff"]
 


MuellerSeb:
Is this necessary for uv? I think pip just picks the available version for the dependency.

I think, this is more crucial:
https://github.com/GeoStat-Framework/GSTools/pull/386/changes#r3334419554

MuellerSeb:
Base on your comment here: https://github.com/GeoStat-Framework/GSTools/pull/386#issuecomment-3019159560

gstools-core 1.1 supports py 3.8. I dont think we need to update the minimal required version of gstools-core for all python version just because 1.3 is the first to support py 3.14

LSchueler:
The problem is that you get a segfault seemingly out of nowhere, if you update to py3.14 and not update gstools-core to v1.3.
We could include Python-version dependent dependencies, but as we don't break anything with requiring the newest gstools-core version, I have the feeling that that is the much easier and also more maintainable alternative.

And no, this has nothing to do with uv.

MuellerSeb:
Ah, ok. But then there should be also a version restriction on gstools-core if it needs to be recompiled for new python versions. Then LGTM

MuellerSeb (approved):
👍

PR #419, opened 2026-08-03
fix(covmodel): ensure positive percentile_scale using bounded brentq

n0228a:
# `fix(covmodel): ensure positive percentile_scale using bounded brentq`

## Problem
`CovModel.percentile_scale(per)` returned negative distance scales (e.g. `-6.215` for `SuperSpherical` with `nu >= 2`). A percentile scale represents physical distance and must be positive ($r \ge 0$).

## Cause
- Correlation functions are even $\rho(-r) = \rho(r)$, making $1 - \rho(r) - \text{per} = 0$ symmetric with roots at $+r_0$ and $-r_0$.
- `percentile_scale` used unconstrained `scipy.optimize.root`.
- For models with sharp drop-offs (`SuperSpherical` with $\nu \ge 2$), flat gradients near $x_0 = 0.9 \cdot \ell$ caused the solver to overshoot across zero and converge onto the mirror root $-r_0$.

## Fix
Replace `scipy.optimize.root` with `scipy.optimize.root_scalar(method="brentq")` bounded on $[0, b]$:
- Constrains the search to $r \ge 0$.
- Dynamically expands upper bound $b$ until $\text{curve}(b) > 0$.
- Guarantees convergence without derivative sensitivity.

## Verification
```python
import gstools as gs
for nu in (0.5, 1.0, 2.0, 5.0):
    m = gs.SuperSpherical(dim=2, nu=nu, len_scale=10.0)
    print(nu, m.percentile_scale(0.9))
```
- **Before**: `nu=2.0 -> -6.215`, `nu=5.0 -> -4.575`
- **After**: `nu=2.0 -> +6.215`, `nu=5.0 -> +4.575`

Test suite: `pytest tests/test_covmodel.py` (7/7 passed).

### DIFF
--- a/src/gstools/covmodel/tools.py
+++ b/src/gstools/covmodel/tools.py
@@ -27,7 +27,7 @@ import warnings
 import numpy as np
 from hankel import SymmetricFourierTransform as SFT
 from scipy import special as sps
-from scipy.optimize import root
+from scipy.optimize import root_scalar
 
 from gstools.tools.geometric import no_of_angles, set_angles, set_anis
 from gstools.tools.misc import list_format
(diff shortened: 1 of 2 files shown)


src/gstools/covmodel/tools.py
```diff
@@ -439,8 +439,12 @@ def percentile_scale(model, per=0.9):
     def curve(x):
         return 1.0 - model.correlation(x) - per
 
-    # take 'per * len_rescaled' as initial guess
-    return root(curve, per * model.len_rescaled)["x"][0]
```
MuellerSeb, line 443: Just a `return abs(root(curve, per * model.len_rescaled)["x"][0])` would be sufficient as the correlation function is symmetric around 0.

n0228a: image
I don't think that fix is enough. There seem to be edge cases, where that root solver just gets stuck.

PR #420, opened 2026-08-11
Make gstools-core the default and gstools-cython optional

LSchueler:
This includes updating the installation instructions.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -107,10 +107,6 @@ tags
 # own stuff
 info/
 
-# Cython generated C code
-*.c
-*.cpp
-
 # generated version file
 src/gstools/_version.py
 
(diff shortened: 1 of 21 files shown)


LSchueler:
We now have a few Python wrapper functions, which don't actually do anything anymore. However, they at least show the default arguments in GSTools' code. Otherwise, you would have to check the gstools-core repo...
I'm not sure if that's worth it to keep the wrapper functions.

MuellerSeb:
I think we can remove the wrapper functions. Defaults are documented in the high-level routines calling these.
Other this looks good to me.

Beside that we now need to get gstools_core onto conda-forge.

LSchueler:
Done.
One thing still to decide: Shall `pip install gstools[rust]` throw an error or should we keep it for a while in case someone has an elaborate and automatized installation script?

We need gstools_core on conda forge before publishing a new version. So, merging this PR should be fine, as long as we keep that in mind.

LSchueler:
Let's keep `pip install gstools[rust]` for backwards compatibility for now. Ready for review.

